PR #20, opened 2015-08-26
Create initial intellij

TheBigBear:
Based on commit https://github.com/saltstack/salt-winrepo/commit/26d70cf466d069d62f90a3a2be05ba957af25f95 by @DavidJFelix

### DIFF
--- a/intellij.sls
+++ b/intellij.sls
@@ -1,0 +1,20 @@
+intellij-community:
+  14.1.4:
+    full_name: 'IntelliJ IDEAS Community Edition 14.1.4'
+    installer: 'https://download.jetbrains.com/idea/ideaIC-14.1.4.exe'
+    install_flags: '/S'
+    uninstaller: '%Program Files (x86)%\JetBrains\IntelliJ IDEA Community Edition 14.1.4\bin\Uninstall.exe'
+    uninstall_flags: '/S'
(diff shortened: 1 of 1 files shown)


DavidJFelix:
Let me in on the secret. What is salt-winrepo-ng? Should I be sending my PRs here?

TheBigBear:
@DavidJFelix what secret? ;-)

The imminent Berryllium rel. (2015.8.x) will allow for jinja2 templated winrepo sls files.

__BUT__ any windows salt minions  <2015.8 will break, hence a new winrepo was created -> winrepo-ng.

Have a peek at the 'new' vagrant sls file under winrepo-ng https://github.com/saltstack/salt-winrepo-ng/blob/master/vagrant.sls .

So, for now we appreciate the winrepo sls files to be created in the "old" winrep (more typing for you - and less typing for me) and we'll create the "new" templated ones from that.

DavidJFelix:
Oh neat. I just assumed winrepo was handled like formulas and this was already allowed.

DavidJFelix:
one more question @TheBigBear, what's the deal with the msi uninstallers with UUIDs in the field name. Like: https://github.com/saltstack/salt-winrepo-ng/blob/master/virtualbox.sls#L7

How are you getting that number?

PR #23, opened 2015-08-27
Add some more old versions. Why not?

DavidJFelix:


TheBigBear:
@DavidJFelix thanks for these. I did merge them in, but ...
Saltstack the 'owners' of ```winrepo``` + ```winrepo-ng``` will probably have to come up with some sort of guideline as to how many to keep. 
With some software it might make sense to keep many, but with others (most?) only the last few usually matter or are actual realistic candidates for installation.

DavidJFelix:
I get your vibe. My rationale behind this was that virtualbox issues way older versions and I know some people have issues with certain versions. Vagrant should be compatible with almost all of them, but I've seen issues with version matchups before.

TheBigBear:
@DavidJFelix fair enough, and I think intel from the ground like yours are exaclty what constitutes good reasons to keep older versions around. Thanks.

DavidJFelix:
@TheBigBear I think there are several considerations to be made:

I think it may also depend on how closely the package follows the semantic version spec. For example: I've included 1.4.0 - 1.4.3, but according to SemVer, all 1.4 versions should be semantically identical to each other. Needing 1.4.0 should be an exception, not a regular occurrence, so we should probably only really have 1.4.3 since it's the newest. In the case of such exception, it might make more sense for somebody to create their own winrepo with 1.4.0 and add it to their winrepo roots. With all that being said -- I have no idea if vagrant follows SemVer this closely. However -- I am sure that python does, and that adding both 2.7.10, 2.7.9, 2.7.8, etc is silly.

One other consideration though. How is updating performed? If you try to keep a year's worth of packages and somebody updates at a year and 1 month, how will the updater find the uninstaller that it needs to remove the old version and replace it with the new version? Homebrew copes with this by properly linking a directory that it controls, but as far as I understand, winrepo does not do this. Having a directory that is controlled by winrepo could be advantageous for a number of reasons, but since we're not currently doing that, I'll save the pro/con for another discussion if somebody decides it might be necessary for removing old versions that winrepo no longer holds.

PR #54, opened 2015-09-11
Pointed salt 2015.8.0 to the -3 build

twangboy:
-3 is the current builds for 2015.8.0
-2 and the original build will be moved to the Archive directory


TheBigBear:
@twangboy windows is really picky and the winrepo version number needs to match what is writtne tot the windows registry.
so this needed '-3' addition in a couple places in the x64 and x86 files. the winrepo-ng file is fine as is.
I have corrected it in https://github.com/saltstack/salt-winrepo/pull/535

PR #69, opened 2015-09-16
Create initial experimental wmf3 

TheBigBear:
This is an experiment and might not work. do NOT MERGE THIS (as yet)


TheBigBear:
This one and the wmf4 and wmf5 won't work until we have installer support for ```wusa``` ```*.msu``` files.

https://github.com/saltstack/salt/issues/27193
https://github.com/saltstack/salt/pull/27224

Once this works we can add powershell v3 v4 and v 5 public preview to any windows version >= win 7, which will give the cool new tools like oneget etc.

PR #112, opened 2015-10-08
Adds Bareos Client (File daemon) 15.2.1 & 14.2.2 to salt-winrepo-ng

neskorodev:


neskorodev:
Looks like pillar data in winrepo sls files still not working.

PR #126, opened 2015-10-30
Curl Update.

kitplummer:
The 7.45.0 version is the one distributed now.


TheBigBear:
@kitplummer thanks, and good catch on the single - double quote issue.

PR #137, opened 2015-11-13
Add missing 'endfor'

rterbush:


TheBigBear:
@rterbush good catch, thanks.

Issue #138, opened 2015-11-13
Program_Files undefined


rterbush:
Seeing this in the logfile on minion:
```
2015-11-13 17:28:42,593 [salt.utils.templates][ERROR   ][1836] Rendering exception occurred :Jinja variable 'Program_Files' is undefined                                                                                                 
```

rterbush:
PR #139 provided.

PR #139, opened 2015-11-13
Fix up assignment variable names.

rterbush:


TheBigBear:
@rterbush this was a copy of another sls file, did this lead to a problem? I thought Windows was not case sensitive? But I am all for 'uniformity' anyway. Thanks for this clean-up.

TheBigBear:
@rterbush sorry I had looked at this in the wrong sequence and missed issue #138 ot this PR

PR #168, opened 2015-12-17
Fix issues with current repo

rterbush:
Really need to find a way to run some testing against commits. These issues break any salt environment that uses the public repo. Probably not a good idea in the first place. Has prodded me to move to private repositories.


TheBigBear:
@rterbush thanks. But I had already fixed the vagrant collision issue, as soon as you added me to the issue. So I can't merge this PR cleanly. Can you remove the vagrant fix , as I can't merge it as is. 
Else I'll add the missing version of nxlog separately.

TheBigBear:
@rterbush, thanks for the report, this has since been fixed by two separate merges. please don't add separate issues into a single PR in the future.
fixed by https://github.com/saltstack/salt-winrepo-ng/commit/d9b11c212fb345725e777066696ff2379d999826
and the 2nd one fixed by https://github.com/saltstack/salt-winrepo-ng/commit/a0ffc25d65b0fb3c2ed858dff06058ceb0cdc687

rterbush:
@TheBigBear sorry I did not get back to you sooner. Would normally pack those changes up in separate PRs but given a busy day, wanted to at least point you in the right direction.

PR #182, opened 2015-12-27
Winscp forloop

gqgunhed:
Combined all versions of WinSCP into one for-loop.


TheBigBear:
@gqgunhed thanks for this.

PR #183, opened 2015-12-27
Wireshark version update

gqgunhed:
Updated versions to 2.0.0 and 1.12.8 (older versions seem to be no longer available).

WinPcap must now be installed separately (tested only for 2.0.0 version).


TheBigBear:
@gqgunhed thanks for this.

PR #184, opened 2015-12-27
Nmap version update 7.01

gqgunhed:
- Updated version
- combined into for-loop
- Fixed typo in uninstaller

### DIFF
--- a/nmap.sls
+++ b/nmap.sls
@@ -1,18 +1,16 @@
 # just 32-bit x86 installer available
 {% if grains['cpuarch'] == 'AMD64' %}
-  {% set PROGRAM_FILES = "%ProgramFiles(x86)%" %}
+    {% set PROGRAM_FILES = "%ProgramFiles(x86)%" %}
 {% else %}
-  {% set PROGRAM_FILES = "%ProgramFiles%" %}
+    {% set PROGRAM_FILES = "%ProgramFiles%" %}
 {% endif %}
 nmap:
-  {% for version in '7.01', '6.47' %}
-  {{ version }}:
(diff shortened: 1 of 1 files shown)


TheBigBear:
@gqgunhed have you tested this? 
In the past the nmap installer did NOT set the version number in the DisplayVersion field in the uninstaller key in the registry, and consequently the winrepo installer only ever found version "Not Found", which makes it impossible to have a sls file for more than one version that actually works.
The 'best' workaround in this case is probably to always have the winrepo installer sls file only ever point to the latest released version of nmap instead.

gqgunhed:
@TheBigBear I tested installing by pkg.install and removing with pkg.purge (without specific version). As far as I remember it is possible that it showed only "Not Found" as "old:" at the time of removing.
So maybe it's better to keep only the latest version in the file.
You have by far more experience with those installers than me :-)

gqgunhed:
@TheBigBear so should I edit the nmap sls to only include the latest version?

TheBigBear:
Yes, please do, I just double checked, and even the most up-to-date version of the 7.x series of nmap still does not set the windows unistall "DisplayVersion" registry key as it should.

nmap-7 x-no-windows-registry-uninstall-displayversion-set

TheBigBear:
have added upstream PR on nmap project 
https://github.com/nmap/nmap/pull/263

gqgunhed:
very cool :-) I used the old 6.47 sls and just replaced 6.47 with 7.01.

TheBigBear:
@gqgunhed thanks. and once the upstream nmap windows installer has added the version number we can then add a few versions into a 'for' loop, as planned.

PR #185, opened 2015-12-27
moved versions into for-loop

gqgunhed:


TheBigBear:
@gqgunhed thanks for this.

PR #190, opened 2015-12-29
new definition for Bandizip installer

gqgunhed:


TheBigBear:
@gqgunhed thnaks for this submissiocn, but I don't think this is correct? 

Pretty sure that if you install this under a 64-bit windows it almost certainly has the un-installer sitting in %ProgramFiles(x86)% and under 32bit windows it would be as you wrote it.

But in tne 'new' winrepo-ng we can now tell those two apart with some simple jinja2 code. 

Please have a look around at other sls definitions in the 'new' winrepo-ng repo, and then update the sls for this file and I will merge it in.

gqgunhed:
@TheBigBear, yes I know about the jinja2 code to set the PROGRAM_FILES accordingly. But the Bandizip installer installs into %ProgramFiles%. See attached the screenshot from my 64bit Win10 machine and the installed Bandizip left of it.
This is why I left out the if-else construct to set PROGRAM_FILES.

bandizip_install

TheBigBear:
@gqgunhed cool, sorry I doubted you.

gqgunhed:
No problem, better safe than sorry. I highly appreciate any advice and questions. There's always something to learn - especially when dealing with Windows ;-)

PR #219, opened 2016-01-04
update adding 3.6.2 to tortoise-hg

Destreyf:
A newer version of tortoise-hg is available.  This adds support for the newer version


TheBigBear:
@Destreyf thanks for this update.

PR #237, opened 2016-01-07
Fix 'full_name' of reader version 10.1.4

breshead:
Somewhere along the line the installer has changed the name of the package in the registry  from 'Adobe Reader 10.1.4' to "Adobe Reader X (10.1.4)"


TheBigBear:
@breshead thanks for this fix.

PR #238, opened 2016-01-09
added 2.0.1 version for wireshark

gqgunhed:


TheBigBear:
@gqgunhed thanks for this

PR #247, opened 2016-01-15
Correct broken URLs for NSClient 0.4 releases.

b18:
It looks like the URLs for NSClient are broken and have been moved to the GitHub page. I updated the 0.4 releases accordingly & tested pushing to a salt-minion.


TheBigBear:
@b18 thanks for this.

PR #264, opened 2016-02-01
Fixed Full name for Xming

twangboy:


TheBigBear:
@twangboy thanks for this. ;-)

PR #266, opened 2016-02-02
Added Influx Capacitor https://github.com/poxet/Influx-Capacitor

Oro:
Hope that the commit is correct here, I have unfortunately not found any documentation how to contribute to winrepo-ng (or winrepo, where this should work just as fine)
Should I also open up a pull request on winrepo, or will this be phased out soon//done automatically?

### DIFF
--- a/influx-capacitor.sls
+++ b/influx-capacitor.sls
@@ -1,0 +1,10 @@
+influx-capacitor:
+  '1.0.15':
+    full_name: 'Influx Capacitor'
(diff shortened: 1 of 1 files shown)


TheBigBear:
@Oro thanks for this. 

It would be nice, but it is not required nor expected, but if you did do a PR for both repos that would be really nice, but I will do this one.  

Not sure when the old winrepo is going to be phased out. But seeing what versions are still in production out there, it might be quite a while. ;-) 

(Disclaimer: I do not speak on behalf of saltstack - this is my my own opinion)

PR #283, opened 2016-02-12
Fix strawberryperl

twangboy:
Added architecture detection
One SLS file for both architectures
Handles the fact that the file version number is different from
the installed version number

Fixes https://github.com/saltstack/zh/issues/488


TheBigBear:
@twangboy I don't understand what you are trying to do? 

The 32-bit and 64-bit versions have different full_name or msi prodict names.

32-bit = "Strawberry Perl" (and NOT "Strawberry Perl (32-bit)"

64-bit = "Strawberry Perl (64-bit)" (in other words the 32-bit version does not mention the bitness in the full_name.

twangboy:
That's exactly why I have you review these things.... I didn't test 32 bit

twangboy:
I'll look some more

twangboy:
@TheBigBear How does it look now? The main reason for the change was the problem with the version in the filename not matching the version in the msi db. I decided to merge them into one file as well, quite badly apparently.

PR #284, opened 2016-02-14
Separate strawberryperl into x64 and x86

twangboy:
Need ability to install 32bit strawberry perl onto 64 bit systems.
https://github.com/saltstack/zh/issues/488


twangboy:
@TheBigBear ^^^

TheBigBear:
> Need ability to install 32bit strawberry perl onto 64 bit systems.
> saltstack/zh#488
@twangboy that link gives me a ```"page not found"``` ```404``` error

twangboy:
@TheBigBear Sorry, we use zenhub to track some issues and I guess they're only available to us. It's just a related issue we're tracking.

PR #285, opened 2016-02-14
Swap strawberryperl version numbers around for ver. 5.20.3002

TheBigBear:
@twangboy The numbers for 5.22.1 were in the right sequence, but the 5.20.3002 ones were reversed.


twangboy:
@TheBigBear My brain hurts. https://github.com/saltstack/salt-winrepo-ng/pull/286

PR #288, opened 2016-02-16
Remove Notepad++ 6.8.9 (removed from site)

johnccfm:
Removed according to https://notepad-plus-plus.org/news/v6.8.9-has-been-removed.html


TheBigBear:
@johnccfm thanks for this. let's hope 6.8.10 will be out soon.

PR #296, opened 2016-02-26
Microsoft build tools

Cryowatt:
Adding Microsoft Build Tool 2015
- For msbuild support

### DIFF
--- a/microsoft-build-tools.sls
+++ b/microsoft-build-tools.sls
@@ -1,0 +1,10 @@
+microsoft-build-tools:
+  '14.0.23107':
+    full_name: 'Microsoft Build Tools 2015'
+    installer: 'https://download.microsoft.com/download/E/E/D/EEDF18A8-4AED-4CE0-BEBE-70A83094FC5A/BuildTools_Full.exe'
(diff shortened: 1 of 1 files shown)


TheBigBear:
@Cryowatt thanks for this. Are you sure they use a single installer to install either the 64 bit or 32 bit build tools? In every other product they have two different installers, one for 32 bit and one for 64 bit? Can you please double check and confirm, before I merge this?

TheBigBear:
@Cryowatt can you please also make the install flag match all other msi installers and use /qn and /norestart , please? Thanks.

Cryowatt:
Yeah, the installer is a bit weird. It actually installs both the 32-bit and 64-bit versions but they seem to be tied together. 64-bit OSes get both installed but they are removed in one shot. This configuration seemed to make the most sense to me.

Cryowatt:
I updated the installer flags.\

TheBigBear:
@Cryowatt, ok, thanks for double checking. Never seen this before from MS, but take your word for it. Thanks for adding this sls file.

PR #300, opened 2016-03-01
New Notepad++ version

msielicki:
There is newer version of Notepad++


TheBigBear:
@msielicki thanks for this.

PR #331, opened 2016-03-09
Adding clamav.sls

ameneau-xetus:


TheBigBear:
@ameneau-xetus can you briefly explain your naming of this pkg? why do you start it off with "xv_" and why do you not reflect that in the name of the sls file?
Also what is the difference or relationship between clamwin and clamav, do you know?

ameneau-xetus:
Thanks for the comments @TheBigBear, the naming of the pkg was an oversight on my part, we were using it internally with salt v2015.5.0 and I forgot to rename it. The pkg also includes both x64 and win32 versions, and we were only using the 64 bit version, so I've removed the _x64 suffix to the pkg name as well.

I'll fix it and create a new pull request.

TheBigBear:
@ameneau-xetus thanks for this. 
Just a comment/note, which you might already know, you don't actually have to use the GUID to uninstall, you can use the saltfs path to the msi installer file alone, in which case you don't actually have to look up the GUID to write the sls msi installer and uninstallers file.

PR #333, opened 2016-03-10
The installer link is broken for version 2.9.1347, removing it. I wou…

ameneau-xetus:
…ld strongly suggest moving to saltfs hosted installer packages to prevent this from happening in the future.


TheBigBear:
@ameneau-xetus thanks for this.

PR #334, opened 2016-03-10
Adds the url added to the nsclient++ installer version 0.4.3.143 as w…

ameneau-xetus:
…as done for saltstack/salt-winrepo#586

### DIFF
--- a/nsclient.sls
+++ b/nsclient.sls
@@ -46,18 +46,3 @@ nsclient:
     msiexec: True
     locale: en_US
     reboot: False
-  '0.3.9.328':
-    {% if grains['cpuarch'] == 'AMD64' %}
-    full_name:  'NSClient++ (x64)'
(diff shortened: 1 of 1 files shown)


ameneau-xetus:
Oh yikes! Thanks for catching that @TheBigBear! I've fixed it and rebased the branch, so it should be good to go!

ameneau-xetus:
hmm... so, I just noticed something testing 32bit installers, the full_name is wrong (should be NSClient++ (x86) instead of NSClient++) and causes the state to fail.

In addition, the link for version 0.3.9.328 is broken

I'll update this to resolve both of those issues unless you (@TheBigBear) recommend having the fixes happen separately

TheBigBear:
No, thats ok, to fix two issues on same package in one PR. Thanks.

ameneau-xetus:
Thanks again for the comments! it should be good to go now

PR #346, opened 2016-03-21
Simplefy, removed deadlink, and added new version

VietTPham:


TheBigBear:
@VietTPham thanks for this

PR #348, opened 2016-03-25
Supporting Firefox install in x64

moebiuseye:
I think this should make it possible for both AMD64 and x86 users to install firefox in their 
respective architecture. 

Here's the result using Windows Server 2008 in AMD64:

(I can test it with Windows Seven Pro AMD64 later today). 
I have no 32 bit windows installation. 

~~~
root@my-salt:~# salt -C 'G@os:Windows' pkg.install firefox
my-windows-server-2008:
    ----------
    Mozilla Firefox 45.0 (x64 en-US):
        ----------
        new:
            45.0
        old:
    Mozilla Maintenance Service:
        ----------
        new:
            45.0
        old:
    _comment:
        Registry not updated.
~~~

### DIFF
--- a/7zip.sls
+++ b/7zip.sls
@@ -14,5 +14,5 @@
     install_flags: '/qn /norestart'
     uninstall_flags: '/qn /norestart'
     msiexec: True
-    locale: en_US
+    locale: {{grains['locale_info']['defaultlanguage']}}
     reboot: False
--- a/7zip_beta.sls
+++ b/7zip_beta.sls
@@ -15,5 +15,5 @@
     uninstall_flags: '/S'
     msiexec: False
     reboot: False
-    locale: en_US
+    locale: {{grains['locale_info']['defaultlanguage']}}
   {% endfor %}
--- a/activeperl_x64.sls
+++ b/activeperl_x64.sls
@@ -7,6 +7,6 @@ activeperl_x64:
     install_flags: '/qn /norestart'
     uninstall_flags: '/qn /norestart'
     msiexec: True
-    locale: en_US
+    locale: {{grains['locale_info']['defaultlanguage']}}
     reboot: False
   {% endfor %}
--- a/activeperl_x86.sls
+++ b/activeperl_x86.sls
@@ -7,6 +7,6 @@ activeperl_x86:
     install_flags: '/qn /norestart'
     uninstall_flags: '/qn /norestart'
     msiexec: True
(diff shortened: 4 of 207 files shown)


twangboy:
This needs to be tested. If I recall we had some problems using %Program Files% before.

moebiuseye:
Yeah, I just figured this isn't just as super-dandy as I thought.
Plus, the registry key couldn't be correctly set.

Is there an issue related to this?

twangboy:
The registry key issue may be fixed depending on the version of salt you're using. It did put it in the registry because it showed a new version. I don't know why it showed the "Registry not updated" thing.

moebiuseye:
I'll make a new pull request later. I think I figured out why salt couldn't find the value in the registry.

TheBigBear:
@moebiuseye I don't understand why you (or anybody?) would want or choose to install the 64-bit version of firefox on a 64 bit version of windows? 

And I can't think and havent heard of a company that would decide to always roll-out 64 bit firefox on 64 bit windows by default?

There is almost no working plugins for firefox 64-bit. The memory consumption for multiple tabs is incredibly wasteful and there is no documented speed or stability improvement. And it has only been out for a few weeks now, and is still not latching on in a big way.

I compare it in many ways to the 64bit version of MS Office that does exist but even Microsft say, we recommend NOT to install it and think it nly makes sense to install it in very specific situations and circumstances.

@twangboy I would quite strongly advise saltstack NOT to switch firefox installs to 64 bit by default, based on the windows OS bitness alone. 

IMHO, the most I would suggest to consider doing would be to allow a explicit and separate 64-bit pkg name such as firefox_x64 AND if people put that in and specifically do want the 64-bit version installed they can do that by choice for the few machines where they expect an advantage or really need or want this.

This would be in-line with the pkg name choice for programming languages such as python, ruby, perl or the MS runtimes, where the admin chooses which bitness to install on a minion.

Here is but a couple background links about the 32 bit vs 64 bit versions of browsers, but I am sure you can get same or better results by using some google foo.

http://arstechnica.co.uk/information-technology/2015/12/firefox-for-windows-finally-has-an-official-stable-64-bit-build/

http://www.ghacks.net/2016/01/03/32-bit-vs-64-bit-browsers-which-version-has-the-edge/

moebiuseye:
I understand now. Thank you for taking the time to point why this isn't a good idea.

PR #351, opened 2016-03-28
Create initial SSC Serv Free 3.6.1

m03:
SSC Serv is a service for Microsoft Windows, which collects system statistics and submits them to a central statistics server. It is similar to and compatible with collectd, a free and open-sourced solution for UNIX systems.

https://ssc-serv.com/features.shtml

Tested install on Windows 2012r2 with salt-minion 2016.3.0.


TheBigBear:
@m03 thanks for this

PR #355, opened 2016-04-06
Update salt-minion.sls

TheBigBear:
@twangboy ^^^ @UtahDave ^^^ is this OK? 

I think the new winrepo-ng windows minion has accidentally been called ```'saltstack.minion'``` instead of the 'standard' expected and documented name of ```'salt-minion'```, right from the set-go out of the gate on the Berryllium release 2015.8.x?

Can we rename this back to what it should be, so it will align with the names used, and expected, in the formulas etc?

I am not going to merge this and potentially break everybodies 2015.8.x setup unless I have the go ahead of at least one of the two of you. ;-)


TheBigBear:
@UtahDave ^^ @twangboy ^^ any comment on this? I think this pkg name needs changing? Do you agree?

twangboy:
@TheBigBear I think you're right. The difficulty lies in not messing up existing setups.... I don't know the best way to do this... which is why I haven't replied.

UtahDave:
Better to fix this now than to get even more people using the incorrect name.

Thanks @TheBigBear

PR #358, opened 2016-04-08
Update JRE8 (add jre8u77)

danielcb:
Add new jre8u77 update.


TheBigBear:
@danielcb thanks for this.

PR #360, opened 2016-04-13
WIP: Initial formula for installing the OwnCloud desktop client

nicoschoenicke:
This installs the ownCloud desktop client.
The uninstaller doesn't work.
If someone has any ideas feel free to let me know.

### DIFF
--- a/owncloud.sls
+++ b/owncloud.sls
@@ -1,0 +1,15 @@
+owncloud:
+  {% if grains['cpuarch'] == 'AMD64' %}
+    {% set PROGRAM_FILES = "%ProgramFiles(x86)%" %}
+  {% else %}
+    {% set PROGRAM_FILES = "%ProgramFiles%" %}
+  {% endif %}
+  '2.2.0.6076':
+    full_name: 'ownCloud'
+    installer: 'https://download.owncloud.com/desktop/stable/ownCloud-2.2.0.6076-setup.exe' 
+    install_flags: '/S'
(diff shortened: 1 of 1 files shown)


TheBigBear:
@nicoschoenicke thanks for this, could you re-work this, so it installs the win 32 version on win32 systems and the win 64 version on win 64 , or if it is the same installer that gets installed on both , at least make sure that the un-installer runs correctly on  the win 32 and the win 64 windows platforms?
You can look at other existing sls files and copy the appropriate relevant way of doing this.

nicoschoenicke:
I made the requested change to the sls file.
Unfortunately this doesn't change my problem of being unable to uninstall the OwnCloud client after an successful installation.

TheBigBear:
@nicoschoenicke ok, i looked into this, and you have one issue caused by two discrepancies. As per documentation, the pkg_fullname needs to match 100% to the name the software puts in the registry.

the ```'ownCloud'``` installer puts a fullname of ```'ownCloud'``` in the registry. Your sls file calls it ```'OwnCloud Desktop Client'```.

Hint, use lowercase ```'o'``` at the start instead, and drop of the  ```' Desktop Client'``` at end and it will work.

Please re-work your file and I will merge it once it is fixed.

Thanks for your contribution.

nicoschoenicke:
I made the changes to the full_name and updated the versions to 2.2.0.6076.

TheBigBear:
@nicoschoenicke thanks for this.

PR #365, opened 2016-04-19
Update openvpn.sls

adnanJP:
Fix bad identation in sls file.


TheBigBear:
@adnanJP thanks for this

PR #368, opened 2016-04-25
Update JRE8 (add jre8u91)

danielcb:


TheBigBear:
@danielcb thanks for this

Issue #373, opened 2016-05-04
Adobe reader dc


hrumph:
Hi all the adobe reader dc links are all dead (I asssume that's normal)
and there's a new one:
https://ardownload2.adobe.com/pub/adobe/reader/win/AcrobatDC/1501020060/AcroRdrDC1501020060_en_US.exe

I think that things just changed at adobe. The version you get isn't

2015.xxx.xxxxx

Instead it has type 15.xxx.xxxx

And explanation of the new versioning stuff is here:

http://www.adobe.com/devnet-docs/acrobatetk/tools/AdminGuide/whatsnewdc.html

The long and short of it is that we probably need to make new sls file called adobe-reader-dc-classic
(or something like that).

It can start off with

```
adobereader-dc-classic:
  '15.010.20060':
    full_name: 'Adobe Acrobat Reader DC'
    installer: 'https://ardownload2.adobe.com/pub/adobe/reader/win/AcrobatDC/1501020060/AcroRdrDC1501020060_en_US.exe'
    install_flags: '/msi EULA_ACCEPT=YES ALLUSERS=1 REMOVE_PREVIOUS=YES /qn'
    uninstaller: 'msiexec.exe'
    uninstall_flags: '/qn /x {AC76BA86-7AD7-1033-7B44-AC0F074E4100} /norestart'
    msiexec: False
    locale: en_US
    reboot: False
```

TheBigBear:
@hrumph you up to doing the new ones and should we then also include the new one and put its URLs in the old sls file? 
You up to making up those two files for us?

hrumph:
@TheBigBear 
sure I'll do a pull request very soon with the new proposed file if you like.
I dont think that the old file should have the new entry because of the way the versioning changed.
15.010.20060 is now the latest version but salt will think that its previous to all of the 2015
versions when it compares on version numbers, so we have to start fresh with a new file.

Issue #352, opened 2016-03-29
Adobe Reader DC download links broken


grandmasterv:
Hello!

It looks like the download locations for Adobe Reader DC have changed.

Old Broken Link
https://ardownload2.adobe.com/pub/adobe/reader/win/AcrobatDC/1501020059/AcroRdrDC1501020059_en_US.exe

New Working Link
https://admdownload.adobe.com/bin/live/readerdc_en_ha_install.exe

https://github.com/saltstack/salt-winrepo-ng/blob/master/adobereader-dc.sls

hrumph:
@grandmasterv 
The link
https://ardownload2.adobe.com/pub/adobe/reader/win/AcrobatDC/1501020060/AcroRdrDC1501020060_en_US.exe

is working. It looks like adobe is purposefully making it's old links dead.

Note that when you use the new link the version of adobe reader dc you get
is 15.xxx.xxxxx
and NOT 2015.xxx.xxxx

This is because adobe has started a new versioning system and the 15.xxx.xxxx versions
are the "classic" track (as opposed to continuous track),

so we need to start fresh with a  new sls file called adobereader-dc-classic

It can start off with
```
  '15.010.20060':
    full_name: 'Adobe Acrobat Reader DC'
    installer: 'https://ardownload2.adobe.com/pub/adobe/reader/win/AcrobatDC/1501020060/AcroRdrDC1501020060_en_US.exe'
    install_flags: '/msi EULA_ACCEPT=YES ALLUSERS=1 REMOVE_PREVIOUS=YES /qn'
    uninstaller: 'msiexec.exe'
    uninstall_flags: '/qn /x {AC76BA86-7AD7-1033-7B44-AC0F074E4100} /norestart'
    msiexec: False
    locale: en_US
    reboot: False
```

hrumph:
I'm closing this. It can be reopoenned if @grandmasterv is dissatisfied with the new merge.

grandmasterv:
Looks good thanks!

PR #374, opened 2016-05-04
New formula for adobereader-dc which is adapted to the new version numbers

hrumph:
WIth the enterprise installer we'ere now getting the "classic" version as ooposed to "continuous", and the version numbers are different so I've made a new sls file.


hrumph:
This should resolve issues #373 and hopefully #352

hrumph:
Sorry I have two different things in the same pull request. Not that familiar with git but the second commit is still good. (The "ESR" string is missing in the firefox-esr.sls file.

TheBigBear:
@hrumph thanks for this

PR #396, opened 2016-05-07
Fix full_name for PostgreSQL

rrzaripov:
Curious, full_name for the PostgreSQL contains a white space in end. Without this fix result of uninstallation will be next:

```
salt win-230416-010535 state.single pkg.removed postgresql
win-230416-010535:
----------
          ID: postgresql
    Function: pkg.removed
      Result: True
     Comment: All specified packages are already absent
     Started: 02:00:58.681000
    Duration: 62.0 ms
     Changes:

Summary for win-230416-010535
------------
Succeeded: 1
Failed:    0
------------
Total states run:     1

```

### DIFF
--- a/postgresql.sls
+++ b/postgresql.sls
@@ -1,6 +1,6 @@
 postgresql:
   '9.5':
-    full_name: 'PostgreSQL 9.5'
+    full_name: 'PostgreSQL 9.5 '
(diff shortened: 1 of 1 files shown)


TheBigBear:
@rrzaripov thanks for this. How bizarre. Should really also be reported upstream, so they can fix this.

PR #474, opened 2016-06-06
Add 1.6.2 installer to golang package.

jbest84:


TheBigBear:
@jbest84 thanks for this.

PR #489, opened 2016-06-15
New vagrant version 1.8.4 released

rrzaripov:
Time to update sls


TheBigBear:
@rrzaripov thanks for this.

PR #510, opened 2016-06-27
Update sensu.sls

jdvachal:
Updated the sensu pkg to use the latest sensu installer

### DIFF
--- a/sensu.sls
+++ b/sensu.sls
@@ -1,5 +1,5 @@
 sensu:
-  '0.21.0.1':
+  '0.25.3.1':
     full_name: 'Sensu'
     installer: 'https://sensu.global.ssl.fastly.net/msi/sensu-0.25.3-1.msi'
     install_flags: '/qn /norestart'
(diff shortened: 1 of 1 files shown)


TheBigBear:
@jdvachal hi thanks for this. is there any reason why you are not using the links at the project, like the earlier versions did?

TheBigBear:
@jdvachal ^^ ???

jdvachal:
No reason.  The fastly link used in this commit was pulled from the reference on the sensu site [here](https://sensuapp.org/docs/latest/platforms/sensu-on-microsoft-windows.html#download-and-install-sensu-core).

I'll gladly update this commit to use a different source for the msi - just need some guidance on from where.

TheBigBear:
@jdvachal ok, seeing that the project itself points to this fastly.net site for it's downloads I think this change is fine. The old format under http://repositories.sensuapp.org/msi/<msi-file-name>.msi also still works.

Just double checked as we try to avoid most of the popular 3rd party download sites as some of them 'mess' with the actual archives and generally it is better following the adivce of any given project. Sorry, shoul have thoguht to go and follow the sensu project links myself, but we are maitaining soo mnay sls files that this becomes quite a chore. Thanks for taking time to report back and submit this update.

So this is absolutely fine. will merge it now.

PR #586, opened 2016-07-19
Update jre8.sls

jbest84:
Bump version to .15 which matches Oracle's latest installer for jre-8u91-windows-x64.exe

### DIFF
--- a/jre8.sls
+++ b/jre8.sls
@@ -1,5 +1,5 @@
 jre8:
-  '8.0.910.14':
+  '8.0.910.15':
     {% if grains['cpuarch'] == 'AMD64' %}
     full_name: 'Java 8 Update 91 (64-bit)'
     installer: 'salt://win/repo-ng/jre8/jre-8u91-windows-x64.exe'


TheBigBear:
@jbest84 thanks for this. sorry I missed that change.

jmacfar:
The installer shows .15, but the windows registry shows .14. Since that is what salt uses, I can't do pkg.remove..

```
$ salt 'w8-cl1*' pkg.latest_version jre8
w8-cl1.salt.local:
    8.0.910.15
$ salt 'w8-cl1*' pkg.install jre8
w8-cl1.salt.local:
    ----------
    Java Auto Updater:
        ----------
        new:
            2.8.91.14
        old:
    jre8:
        ----------
        new:
            8.0.910.14
        old:
$ salt 'w8-cl1*' pkg.remove jre8
w8-cl1.salt.local:
    ----------
    jre8:
        ----------
        current:
            8.0.910.15 not installed
```

jmacfar:
Hm. Looks like the url in the comments is just wrong. Downloaded from oracle, shows .15.

Dont know where the url comes from.

TheBigBear:
@jmacfar @jbest84 so are we good now? or do we need .14 afterall? 
I don't understand the comment about the URL in the command just being wrong? I tried both URLs the 64-bit and 32-bit one and they both work for me and do give me jre8 upd 91?
Please confirm if this is now good or needs reverting .15 back to .14?

jmacfar:
If I download the jre from http://javadl.oracle.com/webapps/download/AutoDL?BundleId=207775 (url in jre8.sls) and install it, it shows up for me as 8.0.910.14. It shows by name as Java 8 Update 91.

If I download jre-8u91-x64 by navigating to Archived versions through a browser and installing manually, it shows as 8.0.911.15. This also shows up as Java 8u91 in control panel.

Please, double check by installing from the URL in jre8.sls to make sure I am not just way off..but does seem to me like the URL is for .14 but downloading from archived versions is .15 ?

So

TheBigBear:
@jbest84 @jmacfar OK, so update 101 is now out and I have done the sls file updates. 

BTW I reverted back to .14 for jre8 upd 91 as that is exactly what you get if you do follow the instructions and download from the manual download through to the bundle id. 
The archive link is not available for a download of the recent or current release and does offer more versions and as can be see by upd 91 also more up to date builds like the jump from build 14 to build 15.

But in looking all this I did also notice that the GUID used for un-installing for jre8 u91 had the wrong GUID in it, it had the 64-bit guid instead of the proper 32-bit guid. also fixed that. 

I assume that @jbest84 's initial problem with u 91 not uninstalling properly was likely related to that typo (actually I simply did not notice the two character difference between the 32 and 64 bit guid and used copy and paste, without thinking) 

thanks for taking the time to reporting the issues I hope it has helped me creating better jre8 sls  files with fewer of the hard to spot GUID 'typos' in the future.

PR #605, opened 2016-07-20
nmap: Remove '"' in uninstaller line

jmacfar:
Trailing " in uninstaller stops nmap from being able to be removed


TheBigBear:
@jmacfar thanks for spotting and fixing this.

PR #750, opened 2016-10-13
Update Vagrant, appended versions 1.8.5 and 1.8.6

rrzaripov:
Versions 1.8.5 and 1.8.6 released


TheBigBear:
@rrzaripov thanks for this

PR #752, opened 2016-10-13
Added new 5.0.x and 5.1.x versions

rrzaripov:
Tested install and remove on Windows 7 64-bit


TheBigBear:
@rrzaripov thanks for this

PR #757, opened 2016-10-14
Added version 9.6

rrzaripov:
Little rework, and note about possible type mismatch when specifying version from command line (yes, same problem as in [issue](https://github.com/saltstack/salt/issues/27433)).


TheBigBear:
@rrzaripov thanks for this

PR #769, opened 2016-10-15
Update owncloud client to 2.2.4

mrueg:


TheBigBear:
@mrueg thanks for this

PR #797, opened 2016-10-20
7-zip 16.04 reports 16.04.00.0 as version

lubyou:
Hi,

7-zip 16.04 reports 16.04.00.0 as version.
Currently installations succeed, but verification fails.

```
----------
          ID: 7zip
    Function: pkg.installed
      Result: False
     Comment: The following packages failed to install/update: 7zip=16.04
     Started: 09:59:43.462000
    Duration: 13100.0 ms
     Changes:
              ----------
              7zip:
                  ----------
                  install status:
                      success

```
screen shot 2016-10-20 at 10 48 47

### DIFF
--- a/7zip.sls
+++ b/7zip.sls
@@ -1,7 +1,7 @@
 # both 32-bit (x86) AND a 64-bit (AMD64) installer available
 {% set PROGRAM_FILES = "%ProgramFiles%" %}
 7zip:
-  '16.04':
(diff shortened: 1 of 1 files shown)


TheBigBear:
@lubyou thanks for this, I have reverted back to the earlier revision that had all the correct version numbers, not quite sure what got me to break all of them.

PR #798, opened 2016-10-20
Jinja2 syntax fixes to make files render correctly in salt.

71mb0:
Just some small syntax adjustments to make jinja2 render those correctly again. 
Hope this helps, let me know if this is not the right form or anything.


TheBigBear:
@71mb0 thanks for this

PR #805, opened 2016-11-02
Update node.js-lts.sls

jbest84:
Node LTS is now 6.x


TheBigBear:
@jbest84 thanks for this. btw i discovered the x86 uninstaller had a long standing typo, fixed that as well, so there is an even newer version available of this now as well.

Issue #874, opened 2017-01-17
duplicate ID errors


danielcb:
For a view weeks now I keep seeing these errors:

```
failed:
        3
    failed_list:
        ----------
        salt-winrepo-ng\duplicati.sls:
            - package 'duplicati', repo data for version number skip_urltest is not defined as a dictionary
        salt-winrepo-ng\intellij-community.sls:
            - Failed to compile 'salt-winrepo-ng\intellij-community.sls': Conflicting ID '1'
        salt-winrepo-ng\intellij-ultimate.sls:
            - Failed to compile 'salt-winrepo-ng\intellij-ultimate.sls': Conflicting ID '1'
```

Is this related to my setup? As far as I can see both packages use different IDs and a for loop with just one version. I'm not sure where the conflicting ID is coming from.

DaveOHenry:
The sls files are malformed which prevents the minions from updating their package database. Looks like the winrepo is not properly tested / maintained. I think I read somewhere that it is not recommended to directly use the winrepo, because it may break things.
I had the same problem some weeks / months ago ("pkg.install" on Windows minions did not work any more) and ended up just removing the sls files.

I now see the winrepo more as a list of examples. Best approach would be to maintain an own list of software where you can always guarantee that the sls files are correct, the installation files are present and that the software versions match your own requirements.

twangboy:
@danielcb In previous versions of Salt if there was an error in one of the .sls files in the winrepo, the entire refresh_db operation would fail. Now it skips the failures and continues building the repo database. This is just the output when there are .sls files with issues... normal behavior.

myii:
I have a site where the master is version `2016.11.2` and the minions are of various versions including `2016.3.3` and `2016.11.2`.

When using `module.run` for `pkg.refresh_db`, for the `2016.3.3` minions (or older), I get:

```salt
          ID: prepare_winrepo
    Function: module.run
        Name: pkg.refresh_db
      Result: True
     Comment: Module function pkg.refresh_db executed
```

However, when running for the `2016.11.2` minions:

```salt
          ID: prepare_winrepo
    Function: module.run
        Name: pkg.refresh_db
      Result: False
     Comment: Module function pkg.refresh_db threw an exception. Exception: Error occurred while generating repo db. Additional info follows:
              
              failed:
                  3
              failed_list:
                  ----------
                  salt-winrepo-ng\duplicati.sls:
[...]

PR #897, opened 2017-03-02
Fix errors in duplicati.sls and intellij-*.sls as discussed in issue #874

myii:
@TheBigBear @moebiuseye Refer back to #874 for full discussion.

### DIFF
--- a/duplicati.sls
+++ b/duplicati.sls
@@ -1,5 +1,4 @@
 duplicati:
-  skip_urltest: True
   '1.3.4':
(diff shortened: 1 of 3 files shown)


damon-atkins:
Suggest you change the title to 
`fix errors in duplicati.sls and  intellij-*.sls`

myii:
@damon-atkins I understand that you don't want the error to be associated with `pkg.refresh_db`, so I've rebased and amended the commit.

PR #905, opened 2017-03-14
Update Putty.sls

mercsniper:
Version 0.68+ moves to a MSI file. Grains for x86/x64 for compatibility and not dealing with setting the program file variable. For loop for new versions can be extended with version numbers. The URL for downloading the file also changed.

### DIFF
--- a/putty.sls
+++ b/putty.sls
@@ -1,6 +1,6 @@
 # just 32-bit x86 installer available
 putty:
-{% for version in '0.68' %}
+{% for version in ['0.68'] %}
   '{{ version }}':
     full_name:  'PuTTY release {{ version }}'
     {% if grains['cpuarch'] == 'AMD64' %}
@@ -16,7 +16,7 @@ putty:
     locale: en_US
     reboot: False
   {% endfor %}
-{% for version in '0.67', '0.66', '0.65', '0.64' %}
+{% for version in ['0.67', '0.66', '0.65', '0.64'] %}
   {% if grains['cpuarch'] == 'AMD64' %}
     {% set PROGRAM_FILES = "%ProgramFiles(x86)%" %}
   {% else %}
--- a/putty.sls
+++ b/putty.sls
@@ -1,11 +1,27 @@
 # just 32-bit x86 installer available
-{% if grains['cpuarch'] == 'AMD64' %}
-    {% set PROGRAM_FILES = "%ProgramFiles(x86)%" %}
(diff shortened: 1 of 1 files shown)


mercsniper:
@TheBigBear Update?

putty.sls
```diff
@@ -1,11 +1,27 @@
 # just 32-bit x86 installer available
-{% if grains['cpuarch'] == 'AMD64' %}
+putty:
+{% for version in '0.68' %}
```
UtahDave: Does `'0.68'` need to be in brackets? I think this may just iterate over the contents of the '0.68' string.

mercsniper:
It will eventually have other versions in it and I did it to "future proof" it like the others.

TheBigBear:
@mercsniper well done, for future proofing it. 
@UtahDave just remember that for pkgs where the "full_name:" does NOT contain a version dependant distinguishing mark, salt winrepo and winrepo-ng can't 'cleanly' handle multiple versions. (at least not last time I checked) So for those windows programs they should only ever have one version in the sls file. I had started with loops for loads of pkgs, but over time learned that it is no good having multiple versions if the "full_name:" can't detect and distinguish them apart. So you will find quite a few by now that still have a loop, but only ever have one version in them.

UtahDave:
Does jinja loop over strings differently than python?
Shouldn't 
`{% for version in '0.68' %}`
be
`{% for version in ['0.68'] %}`

mercsniper:
I just copied what was in the file to differentiate between the exe and msi installers.

UtahDave:
In python that will iterate over each individual character, not '0.68' as a single unit:

```
>>> myvar = '0.68'
>>> for item in myvar:
...     print 'item is {0}'.format(item)
... 
item is 0
item is .
item is 6
item is 8
```

If you make it a list it will iterate like you're expecting:
```
>>> myvar = ['0.68']
>>> type(myvar)
<type 'list'>
>>> for item in myvar:
...     print 'item is {0}'.format(item)
... 
item is 0.68
```

TheBigBear:
@mercsniper you going to put the version strings into a list as shown by @UtahDave ?

mercsniper:
Should've fixed now. I think I changed it properly. Still getting this git
thing down.

TheBigBear:
@mercsniper , thanks, so sorry I somehow missed the update you did 11 days ago.

PR #939, opened 2017-04-07
Fix Jinja syntax error: Remove for loop and adjust variables accordingly

kamilhristov:
I noticed that **pkg.refresh_db** fails with the following error message:

```
ERROR: Error occurred while generating repo db. Additional info follows:

    failed:
        1
    failed_list:
        ----------
        salt-winrepo-ng\bulkrenameutility.sls:
            - Failed to compile 'salt-winrepo-ng\bulkrenameutility.sls': Jinja syntax error: expected token 'end of statement block', got '%'; line 16
              
              ---
              [...]
                  uninstaller: '{{ PROGRAM_FILES }}\Bulk Rename Utility\unins000.exe'
                  uninstall_flags: '/VERYSILENT  /SUPPRESSMSGBOXES /NORESTART /SP-'
                  msiexec: False
                  locale: en_US
                  reboot: False
                {% endfor %    <======================
              ---
    success:
        233
    total:
        234
```
Instead of adding closing curly bracket, I have removed the for loop completely since there is only 1 version.

I checked the other sls files and only 32 are using for loops.

I am wondering which approach is considered better practice when there are more than 1 versions ?

### DIFF
--- a/bulkrenameutility.sls
+++ b/bulkrenameutility.sls
@@ -1,16 +1,14 @@
 bulkrenameutility:
-  {% for version in ['3.0.0.1' ] %}
-  '{{ version }}':
+  '3.0.0.1':
     {% if grains['cpuarch'] == 'AMD64' %}
-    full_name: 'Bulk Rename Utility {{ version }} (64-bit)'
+    full_name: 'Bulk Rename Utility 3.0 (64-bit)'
     {% elif grains['cpuarch'] == 'x86' %}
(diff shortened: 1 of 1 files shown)


UtahDave:
I fixed the problem here https://github.com/saltstack/salt-winrepo-ng/pull/941 without realizing you had fixed  it in this PR.

The loop is nice when there are many versions and not changes needed in the options. It's not required, but I think we'll leave it in for now, unless it continues to cause problems.

Thanks for your PR!

PR #940, opened 2017-04-15
Fix bulkrenameutility recipe

alexproca:

PR #942, opened 2017-04-18
Fix bulkrenameutility recipe

alexproca:
Hi @UtahDave !

My earlier PR #940 was a copy paste from a real environment I patched and contained setting PROGRAM_FILES variable setting too. 

#941 solves just one part of the problem (closing the curly bracket). 

The recipe is still not working because on line 12 
`uninstaller: '{{ PROGRAM_FILES }}\Bulk Rename Utility\unins000.exe'` 
PROGRAM_FILES variable is not defined.

Anyways here is a new PR with the rebased branch :)

### DIFF
--- a/bulkrenameutility.sls
+++ b/bulkrenameutility.sls
@@ -1,3 +1,4 @@
+{% set PROGRAM_FILES = "%ProgramFiles%" %}
 bulkrenameutility:
   {% for version in ['3.0.0.1' ] %}
   '{{ version }}':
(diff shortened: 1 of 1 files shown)


UtahDave:
Thanks, @alexproca.  Good fix.

UtahDave:
I wonder if we should just replace `{{ PROGRAM_FILES }}` with `%ProgramFiles%` since that's really a variable that will work cross platform.

alexproca:
All other files are handling PROGRAM_FILES like that so I respected the pattern. I supposed that someone by mistake deleted first line and the last curly bracket.

PR #961, opened 2017-05-02
Adding rabbitmq and erlang

tully32:
Adding Erlang because it's a dependency of rabbitmq.


tully32:
@TheBigBear added 32 bit erlang for those that want it. Rabbit is 64bit only for the windows installer.

TheBigBear:
@tully32 thanks for this

PR #962, opened 2017-05-03
Add nunit-console package

danielcb:
No x86/x64 packages available, pkg.install/pkg.remove tested locally


danielcb:
@TheBigBear, @UtahDave  can you have a look or is it okay to accept them "unseen" myself if I've tested locally?

TheBigBear:
@danielcb thanks for helping out here, salt is no longer my main focus, so I miss alerts, glad you are stepping in the breach.

LGTM

PR #963, opened 2017-05-03
fixing the application names for rabbitmq and erlang

tully32:
Didn't name them correctly so salt wasn't able to see if they successfully installed.


tully32:
Thanks @TheBigBear

PR #968, opened 2017-05-08
putty.sls not working for me (fixed)

dafyddj:
Putty adds the tag " (64-bit)" in Programs & Features when the 64-bit version is installed.
Also uses the full version number eg. 0.69.0.0 
Previous sls did not match these correctly.


danielcb:
thanks @dafyddj

PR #969, opened 2017-05-11
Update LibreOffice version to 5.2.7.2

akoscomp:
I update version to last Still version.

### DIFF
--- a/libreoffice.sls
+++ b/libreoffice.sls
@@ -1,15 +1,17 @@
 libreoffice:
-  '5.2.7.2':
-    full_name: 'LibreOffice 5.2.7'
+  {% for version in ['5.2.4.2', '5.2.7.2'] %}
+  '{{ version }}':
+    full_name: 'LibreOffice {{ version }}'
     {% if grains['cpuarch'] == 'AMD64' %}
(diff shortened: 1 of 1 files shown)


danielcb:
I think it would be better to add a new version and not change/overwrite a already used version. I'm not sure how states behave, if they use the version you just deleted. @TheBigBear ?

akoscomp:
The updated/deleted version was 5.2.4, and 5.2.7 is a minor update for it, only security and stability issues.
LibreOffice have two active type of installers:
Still (now 5.2.7) - a stable, version, recommended for everyone
Fresh (now 5.3.3) - a new version, is stable, released, but recommended for early adopters and new function testers. In 4-6 weeks this will become a Still version, and the support for 5.2.x will be end.

danielcb:
I don't think anyone should use/stay at the old verison, I'm just not sure what happens if your state depends on a version that isn't available anymore in the repo.

akoscomp:
Now I understand, what do you want. I'm new in salt, and I don't know that is important to keep the old version. I make a new SLS file, and please drop this request.

danielcb:
Thanks for updating and sorry for the delay.

PR #970, opened 2017-05-17
Fixing the 7zip 9.20 download locations

tully32:
Old url's were broken. Needed to be updated.


danielcb:
They seem to host all versions on their project page also. And they seem to be using more or less generic download links: http://www.7-zip.org/a/7z920.msi

Should we maybe add those instead?

tully32:
You're correct. Updated to reflect the proper urls.

TheBigBear:
@tully32 thanks for those updates

PR #980, opened 2017-06-09
7zip improvements

fizmat:
- Rewrite 7zip.sls with templating to have less repeating code.
- Use https to download installers.
- Add parameter "arch" that controls which msiexec is used.


TheBigBear:
@fizmat thanks for this. what tests have you performed on this ?

PR #981, opened 2017-06-09
add muCommander

fizmat:


TheBigBear:
@fizmat this syntax won't work for sure. please test and then re-submit working code. jinja variables are enclosed in double squirly brackets {{ varname }}.

twangboy:
@fizmat Please fix this to work on both 32 and 64 bit systems so we can get this in.

TheBigBear:
@twangboy @fizmat have you tested this? last time I did some checking one could not have multiple versions, if they did not reflect a change in the full_name, has the win salt pkg handler grown up and can now handle matching that correctly?

twangboy:
@TheBigBear I think it checks the version entry in the registry as well. The full name has to match exactly what's in `Add/Remove Programs` and the version has to match as well.

The use of `Program Files (x86)` leads me to believe this will only function on 64bit operating systems.

TheBigBear:
@fizmat this sls will not work 'as-is' on a win 32 bit os as it uses the ```Program Files (x86)``` path, which won't exist there. 
use a jinja variable instead to make it work on either

PR #985, opened 2017-06-15
Download online instead of manually place file on master

sstodd7532:


TheBigBear:
@sstodd7532 thanks for this.

PR #988, opened 2017-06-16
Update to use different URL

sstodd7532:
The previous URL for FossHub did not work.


TheBigBear:
@sstodd7532 thanks for checking and correcting this

PR #993, opened 2017-06-24
Git: add 2.13.1.2 and consolidate versions using jinja

dafyddj:
Hand tested and pushed out to production.


TheBigBear:
@dafyddj Neat. Thanks for this.

PR #1057, opened 2017-09-13
Update npp to v7.5.1

sstodd7532:

PR #1097, opened 2017-11-08
Added new NPP version

couilllard45682:


couilllard45682:
Dup of #1057 but with the correct fix.

PR #1133, opened 2018-01-04
Added scale out state server

kaluznyt:


TheBigBear:
@kaluznyt thanks for providing this PR

PR #1134, opened 2018-01-04
updated urlrewrite

kaluznyt:


TheBigBear:
@kaluznyt have you tested this? 

Last time I checked winrepo-ng could not properly handle detection of a version being installed or upgraded if the 'full_name' of the software (which you have no control over) does not reflect the version of change between different versions.

Can you please check this works as you expect? Things to check are install the older lower version and see if upgrade works, then the other way around will it do the right thing? Finally can you 'force' a version and get winrepo-ng do the right thing?

If you find it still can't handle properly recognising and differentiating between the installed versions you might have to reduce the versions in the sls file to just one for the latest version, maybe?

TheBigBear:
@kaluznyt thanks for doing these tests, looks like there has been progress in winrepo-ng land that I missed. will merge your PR, I think downgrade is not always something that windows software installs cleanly, so that failure is ok for now.
Thanks for your contributions.

PR #1135, opened 2018-01-04
added queue explorer pro

kaluznyt:


TheBigBear:
@kaluznyt this sls seems to assume a 64 bit OS install? ( hard coded x86 program files folder for uninstall )

Please re-write using jinja to help do the right thing regardless of 32 bit or 64 bit windows install.

kaluznyt:
Actually I think there is only 32 bit version of the software

TheBigBear:
that may be the case, but if installed on a 32 bit windows the uninstaller in your sls will fail. 

so you have to always write for both, copy from another sls that is already done right.

kaluznyt:
should be fine now ?

TheBigBear:
@kaluznyt thanks for your PR

PR #1164, opened 2018-01-10
Update winscp.sls

airstream:


TheBigBear:
@airstream thanks for the PR

PR #1184, opened 2018-01-12
Fix the lazarus software definition file

twangboy:
Fixes the problems I merged with the lazarus software definition file.
Adds a note about possible issues with lazarus installation.

### DIFF
--- a/lazarus.sls
+++ b/lazarus.sls
@@ -1,3 +1,14 @@
+# NOTE: If the lazarus installation directory exists the installation will
+#       pop up a dialog box. This will cause the minion to hang because it
+#       is waiting for the user to click continue on a dialog box that will
+#       never be seen. So, before running this, make sure the lazarus
+#       installation directory is not present.
+#       If doing an upgrade that means you'll have to do something like the
+#       following:
+#       pkg.remove lazarus
+#       file.remove c:\lazarus
+#       pkg.install lazarus
+
 {% set versions = [('1.8.0', '3.0.4'),
                    ('1.6.4', '3.0.2')] %}
 lazarus:
--- a/lazarus.sls
+++ b/lazarus.sls
@@ -1,16 +1,15 @@
+{% set versions = [('1.8.0', '3.0.4'),
+                   ('1.6.4', '3.0.2')] %}
 lazarus:
-  1.8.0:
-    full_name:
-      Lazarus 1.8.0
-    installer:
-{%- if salt.grains.get('cpuarch') == 'AMD64' %}
-      https://gigenet.dl.sourceforge.net/project/lazarus/Lazarus%20Windows%2064%20bits/Lazarus%201.8.0/lazarus-1.8.0-fpc-3.0.4-win64.exe
-{%- else %}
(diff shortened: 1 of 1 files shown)


TheBigBear:
@twangboy LGTM will merge it. 

I have one question in principle though, is it good practice and more efficient to just assume that if it is not ```AMD64``` cpuarch it is going to be ```x86```, and only use a ```else``` or would it be better to check the cpuarch with an other ```elif``` for ```x86``` instead, that way nothing breaks if ```itanium``` or something else comes around as supported by salt. Is there a speed or efficiency loss with adding an explicit and (currently) superfluous cpuarch check at that stage?
I noticed that I am inconsistent across the many sls files I have touched and sometimes I do it one way, but not always? Your thoughts, sir?

TheBigBear:
@twangboy ok not sure what is wrong and not sure who to point the finger at, now? 
Sorry for having suspected your merge of the lazarus sls file earlier.

So it wasn't the 'lazarus' software definition that was the culprit for the faults I saw. But 'lazarus.sls'  just happended to be a sls file that was recently changed, sorry.

I have since stepped back in the repo and not yet found where or how the repo-ng broke and stopped compiling.

This is the error I get when I run a ```pkg.refresh_db``` for any and all of my minions

```
    The minion function caused an exception: Traceback (most recent call last):
      File "c:\salt\bin\lib\site-packages\salt\minion.py", line 1470, in _thread_return
        return_data = executor.execute()
      File "c:\salt\bin\lib\site-packages\salt\executors\direct_call.py", line 28, in execute
        return self.func(*self.args, **self.kwargs)
      File "c:\salt\bin\lib\site-packages\salt\modules\win_pkg.py", line 605, in refresh_db
        return genrepo(saltenv=saltenv, verbose=verbose, failhard=failhard)
      File "c:\salt\bin\lib\site-packages\salt\modules\win_pkg.py", line 761, in genrepo
        successful_verbose
      File "c:\salt\bin\lib\site-packages\salt\modules\win_pkg.py", line 829, in _repo_process_pkg_sls
        for pkgname, versions in six.iteritems(config):
      File "c:\salt\bin\lib\site-packages\salt\ext\six.py", line 606, in iteritems
        return d.iteritems(**kw)
    AttributeError: 'str' object has no attribute 'iteritems'
```

I know this is NOT the place to raise this issue. And I will raise a proper issue on the main salt repo over the weekend. But I thought you might know how to find the real culprit and be able to fix it? Maybe even before I wake up for the weekend?

twangboy:
@TheBigBear 
With regards to using `else` or `elif`, I would imagine the `elif` uses more cycles as it's an additional if statement. I've always used else myself. And as far as salt on Itanium... well, I'm not creating Itanium packages for salt. I dont' think we have a stance on that right now. In fact, Python doesn't have an installer for itanium, so I would say just use `else` for now. If itanium suddenly resurges in popularity, then we'll address the issue.

With regards to me messing up the software definition file for lazarus... I did mess it up. And I merged it without testing it. There were many problems even with the initial package submitted by the user that would not have worked... which I merged... I take full responsibility on that one.

The stack trace you were seeing above should be taken care of by: https://github.com/saltstack/salt/pull/45467. It doesn't fix the stack trace itself, but fixes the issue with caching files from hidden directories such as `.git`

Thanks for staying on top of winrepo. I appreciate the work you're doing.

PR #1186, opened 2018-01-13
Create pgadmin4.sls

airstream:
pgAdmin 4 software
https://www.pgadmin.org/

### DIFF
--- a/pgadmin4.sls
+++ b/pgadmin4.sls
@@ -1,0 +1,19 @@
+pgadmin4:
+{% for package_version in ['2.1',
+                           '2.0',
+                           '1.6'] %}
+  '{{ package_version }}':
+    {% set PROGRAM_FILES = "%ProgramFiles(x86)%" %}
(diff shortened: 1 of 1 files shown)


TheBigBear:
@airstream thanks for this. this sls file will not work right on 32 bit OS installs. Please use ```cpuarch``` checks and make it work on either a 32 or 64 bit install.

airstream:
@TheBigBear can you please test a new commit of pgadmin4.sls state? 
Unofrtunately I don't have any Windows 32bit OS computer. Pgadmin4 software installation file exists for x86 https://www.postgresql.org/ftp/pgadmin/pgadmin4/v2.1/windows/

TheBigBear:
@airstream LGTM, thanks 'normally' all that is required if the installer is already only x86 32-bit, is to tweak the uninstaller Program Files location to be correct. You have done that here, so I will merge it, without further tests on my side.  Thanks.

PR #1188, opened 2018-01-14
Create soapui.sls

airstream:
SoapUI Open Source https://www.soapui.org/downloads/soapui.html

### DIFF
--- a/soapui.sls
+++ b/soapui.sls
@@ -1,0 +1,17 @@
+soapui:
+{% for version in ['5.4.0'] %}
+  '{{ version }}':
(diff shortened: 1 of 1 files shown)


TheBigBear:
@airstream thanks for this PR. merging it as-is 
q: is this software really showing the version number twice in the full name in the program add/remove display string? ( or was that a copy / paste slippage ? )

airstream:
@TheBigBear yes, soapui software shows the version number twice in the full name.
Screenshot was added.
soapui

TheBigBear:
Thanks for confirming this. I do not install al those packages, and this just looked odd, but that is fine then.

PR #1192, opened 2018-01-14
Update vlc.sls

airstream:
Fixed.


TheBigBear:
@airstream thanks for this quick fix and reaction, well done

PR #1251, opened 2018-01-20
Change audacity.sls syntax on branch Travis ci cron daily

TheBigBear:


audacity.sls
```diff
@@ -1,33 +1,18 @@
+{% set versions = ['2.2.1', '2.1.0', '2.0.6'] %}
```
twangboy: 2.2.1 isn't available on sourceforge. It's on fosshub now? The site sends you to:
https://www.fosshub.com/Audacity.html/audacity-win-2.2.1.exe

twangboy: Might need two loops... One for 2.1.2 and greater and another for the older versions

PR #1292, opened 2018-02-08
Create nomacs.sls

airstream:
https://nomacs.org/

### DIFF
--- a/nomacs.sls
+++ b/nomacs.sls
@@ -1,5 +1,5 @@
-{% for version in ['3.8.0'] %}
 nomacs:
+{% for version in ['3.8.0'] %}
   '{{ version }}':
     full_name: 'nomacs - Image Lounge'
     installer: 'salt://win/repo-ng/nomacs/nomacs-setup-x64.msi
--- a/nomacs.sls
+++ b/nomacs.sls
@@ -1,0 +1,12 @@
+{% for version in ['3.8.0'] %}
+nomacs:
+  '{{ version }}':
+    full_name: 'nomacs - Image Lounge'
+    installer: 'salt://win/repo-ng/nomacs/nomacs-setup-x64.msi
+    install_flags: '/qn /norestart'
+    uninstaller: 'salt://win/repo-ng/nomacs/nomacs-setup-x64.msi
+    uninstall_flags: '/qn /norestart'
+    msiexec: True
+    locale: en_US
+    reboot: False
+{% endfor %}


airstream:
Can you please commit this pull request (nomacs.sls)?

nomacs.sls
```diff
@@ -0,0 +1,12 @@
+{% for version in ['3.8.0'] %}
+nomacs:
```
twangboy: this line (`'nomacs:`) needs to be moved outside the for loop

airstream:
@twangboy done! Can you please explain why nomacs needs to be outside the for loop?

twangboy:
@airstream You used a for loop in your package definition file. It will work correctly as it was because there was a single version in your list. But when you add another version, you will get a repeated top level `nomacs`. That top level item has to be unique. Not only throughout the document, but also throughout winrepo. 

However, I'm seeing a couple issues with nomacs itself.

First of all, since this is going onto the community winrepo, we should probably point the installer to the actual download location instead of a location on the salt file roots. So, the installer value should probably be `http://download.nomacs.org/nomacs-setup-x64.msi`

The second issue I see is a problem with the way nomacs does their downloads. It doesn't look like you can download an older version of the nomacs installer. Google Chrome works in much the same way. You basically have a single download location for the installer which is always the latest version. You have no option to download an older version of the software. So, if you look at the package definition file for chrome [here](https://github.com/saltstack/salt-winrepo-ng/blob/master/chrome.sls) you'll see that instead of having the actual version number, we've used `latest`. That just means that we don't know what version we're downloading, but that it is the latest version available.

The problem with this approach is that the software will be installed each time you do `pkg.install chrome` because Salt has no way of knowing if the local version installed is the same as the version available from Google. There's really no way around it, aside from saving your own copies of the chrome installer with the version number in the file name and maintaining your own package definition file that points to those versions. But in the case of chrome, that's ridiculous as chrome automatically updates itself to the latest version anyway. I don't know if nomacs automatically updates itself, so it may not be as ridiculous if you decided to go that route.

I thought chrome was the only piece of software that worked this way, but you have now introduced me to another.

So, I would model this after the package definition for chrome but add handling for the 32bit version of the installer since nomacs provides one:
```
nomacs:
  latest:
    full_name: 'nomacs - Image Lounge'
    installer: 'http://download.nomacs.org/nomacs-setup-x64.msi'
    uninstaller: 'http://download.nomacs.org/nomacs-setup-x64.msi'
    {% else %}
    installer: 'http://download.nomacs.org/nomacs-setup-x86.msi'
    uninstaller: 'http://download.nomacs.org/nomacs-setup-x86.msi'
    {% endif %}
    install_flags: '/qn /norestart'
    uninstall_flags: '/qn /norestart'
    msiexec: True
    locale: en_US
    reboot: False
```
@TheBigBear what do you think?

airstream:
@TheBigBear what do you think?

airstream:
@twangboy @TheBigBear is it possible to use installation .msi file in a salt repository using `installer: salt://` for state nomacs.sls?
For example curl.sls state use this construction for installer: `salt://win/repo-ng/curl/curl-7.46.0-win64.msi`
https://github.com/saltstack/salt-winrepo-ng/blob/master/curl.sls

airstream:
@twangboy @TheBigBear any news?

TheBigBear:
@airstream sorry I have no opinion on the using the 'latest' string instead of a version, don't know enough about internals of this project to know if this is a good use or a bad idea ... 
as far as using a local salt:// path as the installer path, that certainly does work ... but I cannot readily make the connection between that 'odd' question and this issue? 
Sorry

TheBigBear:
@twangboy can you advise @airstream on what to do, or not to do, on this one?

twangboy:
You can use `salt://path` as the installer path, but you would need to manage those yourself and not push them up to the winrepo here. The `curl.sls` (and any others that exist) should probably be updated to point to a public URL. These are for general usage and should work for everyone connected to the internet.

If I recall, someone had an idea to make it more universal, like check locally first and then reach out to the World Wide Web if it didn't exist locally, but I don't think that's been implemented. I can't remember where the idea came from.

TheBigBear:
@airstream thanks for this PR

TheBigBear:
@airstream sorry, but yoru code has an issue:
```
failed_list:
    ----------
    salt-winrepo-ng\nomacs.sls:
        - Failed to compile 'salt-winrepo-ng\nomacs.sls': while parsing a block
mapping
            in "<string>", line 4, column 5:
                  full_name: 'nomacs - Image Lounge'
                  ^
          expected <block end>, but found '<scalar>'
            in "<string>", line 6, column 21:
                  install_flags: '/qn /norestart'
                                  ^
```
Please review and fix.

PR #1309, opened 2018-02-20
Update evernote.sls to ver. 6.9.7.6770

TheBigBear:


evernote.sls
```diff
@@ -1,4 +1,13 @@
-evernote:
+evernote: 6.9.7.6770
```
robchrob: It seems like a bug?

PR #1311, opened 2018-02-22
fix typo on first line of evernote definition

vernondcole:
Unfortunately, I cannot test this due to the "feature" that one cannot override the default for "winrepo_remotes_ng".  We need to re-open that issue, I think.

### DIFF
--- a/evernote.sls
+++ b/evernote.sls
@@ -1,4 +1,4 @@
-evernote: 6.9.7.6770
+evernote:
   '6.9.7.6770':
     full_name: 'Evernote v. 6.9.7'
     installer: 'https://cdn1.evernote.com/win6/public/Evernote_6.9.7.6770.exe'


vernondcole:
@TheBigBear please review this.

TheBigBear:
@vernondcole so sorry about this. More 'haste' less speed ... ;-)

vernondcole:
Thanks for the quick work.  I discovered last night why I could not switch
to my fork to test my fix.  It is a design error, but they swept it under
the rug three years ago with a documentation change telling us how we can
not use the facility at all.  Not the correct answer. I think I will
re-file the bug report about not being able to override the default URL for
the GitHub repo

Again, thank you.

TheBigBear:
@vernondcole that is interesting? I am using my own github repo as the default in my setup, with no issue, do you mind showing me your report that was 'swept under the carpet'?

PR #1314, opened 2018-02-22
comment that newrelic-infra uninstall doesn't work.

nbirnel:
NewRelic Infrastructure msi uninstall is broken. It would be nice to work around this, but at least this way it is documented.


TheBigBear:
@nbirnel thanks for the PR

PR #1318, opened 2018-02-28
Add a winrepo-ng package definition for LiveScribe Echo Desktop

xenophonf:


TheBigBear:
@xenophonf the version number should be in single quotes - as per docs - can you please change that?

PR #1331, opened 2018-03-02
Update rocketchat

gabrielloberg:
Update rocketchat


TheBigBear:
@gabrielloberg thanks

PR #1332, opened 2018-03-13
Create eraser.sls

airstream:
Software URL https://eraser.heidi.ie/

### DIFF
--- a/eraser.sls
+++ b/eraser.sls
@@ -1,0 +1,14 @@
+eraser:
+{% for version in ['6.2.2982'] %}
+{% for version_full in ['6.2.0.2982'] %}
+  '{{ version }}':
+    full_name: 'Eraser {{ version_full }}'
+    installer: 'https://10gbps-io.dl.sourceforge.net/project/eraser/Eraser%206/6.2/Eraser%206.2.0.2982.exe'
+    install_flags: '--quiet /verysilent /norestart'
+    uninstaller: 'msiexec.exe'
+    uninstall_flags: '/qn /x {DFCF78CC-3DAD-4C1E-8BC6-94DC5B73461E} /norestart'
+    msiexec: False
+    locale: en_US
+    reboot: False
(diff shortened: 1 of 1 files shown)


TheBigBear:
@airstream you are setting yaml variables version and version_full but do not use them? this way the yaml for loops make no sense. can you re-write this?

twangboy:
@airstream The way you have it written, all versions in `version_full` will be applied to each `version`. What you want is where each `version_full` applies to a specific `version`. You need a list of tuples. Then for each new version that comes out, you just add another tuple. Something like this should work:
```
{% set versions = [
    ('6.2.2982', '6.2.0.2982'),
    ('6.2.2979', '6.2.0.2979')
] %}
eraser:
{% for version, version_full in versions %}
  '{{ version }}':
    full_name: 'Eraser {{ version_full }}'
    installer: 'https://downloads.sourceforge.net/project/eraser/Eraser%20{{ version[:1] }}/{{ version[:3] }}/Eraser%20{{ version_full }}.exe'
    install_flags: '--quiet /verysilent /norestart'
    uninstaller: 'msiexec.exe'
    uninstall_flags: '/qn /x {DFCF78CC-3DAD-4C1E-8BC6-94DC5B73461E} /norestart'
    msiexec: False
    locale: en_US
    reboot: False
{% endfor %}
```
Notice that you need more than just `version` and `version_full` for this package. If you look at the URL to the installer there an `Eraser 6` folder, then a `6.2` folder, and then the installer. To future proof this you also need to slice the version to get the major and minor versions.

TheBigBear:
@akoscomp please re-read instructions from @twangboy https://github.com/saltstack/salt-winrepo-ng/pull/1332#issuecomment-374778544 your sls still doesn't make sense nor is it future proof. you should use the defined yaml variables in the installer line and you should introduce the overall structure as suggested by @twangboy

PR #1356, opened 2018-04-04
Added Nextcloud

gabrielloberg:
Added Nextcloud


TheBigBear:
@gabrielloberg this sls as-is will only work for installing on a 64-bit windows. the program files path is wrong for installing it on a 32-bit windows, please look at other sls files here and adapt it to cater for both scenarios.

PR #1365, opened 2018-05-03
Remove unsupported versions of salt

twangboy:
Versions older than 2016.11 are no longer supported and may be unsafe to use
Version 2016.11 is in CVE only support


TheBigBear:
@twangboy should I merge this? or is it waiting on some saltstack internal 'approval' of some sort?

twangboy:
Please merge. Thanks for all you do!

PR #1386, opened 2018-05-30
Fix windows sensu version for different windows versions

ErmalK:


TheBigBear:
@ErmalK thanks for this PR

PR #1387, opened 2018-05-30

stratusjerry:
Fix for multiple CVEs https://helpx.adobe.com/security/products/acrobat/apsb18-09.html


TheBigBear:
@stratusjerry thanks for this PR

PR #1413, opened 2018-06-14
Update nxlog.sls

airstream:


TheBigBear:
@airstream thanks for this PR, do you want to consider re-writing this for the next upgrade and introduce the yaml variables for %version% etc and save yourself some lines with each new future version?

PR #1439, opened 2018-06-26
 enable multiversioning for nxlog

ErmalK:


TheBigBear:
@ErmalK thanks for the PR

PR #1451, opened 2018-07-10

stratusjerry:
Security Update https://helpx.adobe.com/security/products/acrobat/apsb18-21.html


TheBigBear:
@stratusjerry thanks for the PR

PR #1465, opened 2018-07-31
Fix Erlang version detection and add version 21.0.1

bradleyjokinen:
This adjusts the Erlang package to use an installer script that adds an entry for the version to the registry. It also fixes the full name of the package and adds support for the latest Erlang release.


TheBigBear:
@bradleyjokinen thanks for this great PR containing this work-around. Are you sure and confident the windows registry key is being set in the right location in both cases 64 bit and 32 bit windows? I wild have expected the need for a windows WOW6432Node registry reflection workaround as well to set the uninstall key in the right place? Maybe I am wrong, in this case?

bradleyjokinen:
You were right, there were issues with version detection and package removal on 64-bit Windows. I pushed a couple more changes after testing them on both 32-bit and 64-bit Windows.

TheBigBear:
@bradleyjokinen thank you so much for your PR and for testing this. Merging it now.

PR #1472, opened 2018-08-08
Update Python 2 definitions

prensing:
Update the Python 2 definitions to include the lastest versions.

I switched the code to use a Jinja loop over the version numbers, so it is a lot cleaner.  I did not include the Python 3 definitions; while those are also well behind, the format is more complicated, so needs a bit more work.


TheBigBear:
@prensing sing thank you so much for doing this, and finally making use of jijnja in this one too. Will merge it now.

PR #1476, opened 2018-08-10
Add RabbitMQ version 3.7.7

bradleyjokinen:
This adds RabbitMQ 3.7.7 which is compatible with the Erlang 21.0.1.


TheBigBear:
@bradleyjokinen thanks for this PR

PR #1479, opened 2018-08-14
Create Firefox x64 definition

prensing:
New SLS file to add Firefox x64. I did not verify all the old versions, so some might not exist in 64-bit.


TheBigBear:
@prensing have removed versions earlier than 42.0, as that is when the introduced 64 bit versions. 

But the real question is whether it should indeed get it's own pkg name as in 'firefox_x64' as you suggested, or if it should be simply 'firefox' and always install 32 bit on 32 bit and 64 bit on 64 bit ..., but then it would need detection logic for cases where there is already a 32 bit version installed on 64 bit os, and possibly a way to over-ride the choice, to allow users to opt for 32 bit install even on a 64 bit os. 
I think this needs pondering and confirming by @twangboy before we could merge this, for an important and popular pkg such as this one.  
And if we go for firefox_x64, the logical other question is whether the 'old' existing package that sofar always installed the 32 bit version everywhere, now needs renaming to firefox_x86 ? 
Over to you @twangboy

prensing:
Yes, I thought about that a little, but did not want to make that call. 

One thing to keep in mind: I am seeing lots of places on our machines where a package is installed but win_pkg does not pick it up. There are different reasons (eg. missing version number). If you go with just a single package which matches the OS, then some machines will not have it matched because the 32-bit is installed on 64-bit OS.

stratusjerry:
So a cautionary tail, I created 32/64/esr-32/esr-64 repos and ran into issues when users had installed 64bit into the default 32bit directory.  I'm not sure if this was intended or the result of 32 bit being installed and an upgrade either by the user or the builtin firefox upgrade to 64 retaining the 32bit directory path.  The result was that more than one version of 64 bit was showing as installed (via win_pkg.list_pkgs) but only the newest version was able to be uninstalled because the install path of the older version (defined in the sls) was wrong.  To rectify I ended up writing a custom salt state, module and powershell script to detect and remove duplicate installs taking into account the different directories.

prensing:
Actually, I have exactly that problem with ~4 machines, but did not know the cause. But I think it can be fixed manually in the Windows GUI (still to be confirmed).

firefox_x64.sls
```diff
@@ -0,0 +1,14 @@
...
+  '{{ version }}':
+    full_name: 'Mozilla Firefox {{ version }} (x64 en-US)'
+    installer: 'https://download-installer.cdn.mozilla.net/pub/firefox/releases/{{ version }}/win64/en-US/Firefox%20Setup%20{{ version }}.exe'
+    install_flags: '/s'
+    uninstaller: '{{ PROGRAM_FILES }}\Mozilla Firefox\uninstall\helper.exe'
+    uninstall_flags: '/S'
+    msiexec: False
+    locale: en_US
+    reboot: False
+  {% endfor %}
```
twangboy: Since this will only install on a 64bit machine, I recommend wrapping the whole thing in an if statement:
```
{% set versions = ['61.0.2', '61.0.1', '61.0', '60.0.2', '60.0.1', '60.0', '59.0.3', '59.0.2', '59.0.1', '58.0.2', '58.0.1', '58.0', '57.0', '57.0.4', '57.0.3', '57.0.2', '57.0.1', '56.0', '56.0.2', '56.0.1', '55.0', '55.0.3', '55.0.1', '54.0', '54.0.1', '53.0', '53.0.3', '52.0', '52.0.2', '52.0.1', '51.0', '51.0.1', '50.1.0', '50.0', '50.0.2', '49.0', '49.0.2', '49.0.1', '48.0', '48.0.2', '47.0', '47.0.1', '46.0', '46.0.1', '45.0', '45.0.2', '45.0.1', '44.0', '44.0.2', '44.0.1', '43.0', '43.0.4', '43.0.3', '43.0.2', '43.0.1', '42.0'] %}
firefox_x64:
  {% for version in versions %}
...
  {% endfor %}
{% endif %}
```

stratusjerry: You can also remove the line 
`{% set PROGRAM_FILES = "%ProgramFiles%" %}`

and change this line to 
`uninstaller: '%ProgramFiles%\Mozilla Firefox\uninstall\helper.exe'`

PR #1480, opened 2018-08-15

stratusjerry:
Update to newest https://helpx.adobe.com/security/products/acrobat/apsb18-29.html


TheBigBear:
@stratusjerry thanks for this PR

PR #1485, opened 2018-08-30
Fix and standardize urls to sourceforge

twangboy:
Sourceforge allows companies to mirror their downloads so you can get a faster download depending on the region you're in. Each mirror has its own subnet. For example to download a file hosted by IWeb the URL would be something like this `http://iweb.dl.sourceforge.net/project/yourproject/installer.exe`. Mirrors seem to come and go and a mirror you specify in an SLS file may not be available tomorrow. Sourceforge seems to have a standard/primary that will auto select a mirror. Which is just to use `http://downloads.sourceforge.net/project/yourproject/installer.exe`

This modifies all SourceForge links to use `downloads.sourceforge.net` as the download URL.
Add .gitignore


twangboy:
My bad... Thanks for fixing that.

PR #1486, opened 2018-08-31
Create graylog-collector-sidecar.sls

pauldalewilliams:
This PR adds a state file for Graylog Collector Sidecar.  Tested on Windows Server 2016.


pauldalewilliams:
May be better to rename this collector-sidecar as that's the name they give the Linux pkgs:  https://github.com/Graylog2/collector-sidecar/releases/tag/0.1.6

Though it appears they're going with graylog-sidecar with the newer version they're working on:  https://github.com/Graylog2/collector-sidecar/releases

PR #1507, opened 2018-10-24
Url fix

stratusjerry:
Small URL fix and move more packages to supported HTTPS URLs


twangboy:
Thanks for those fixes!

PR #1508, opened 2018-10-28
Updated CPU-Z

samjhandley:
Updated CPU-Z:
  - Use url instead of salt hosted exe
  - Includes latest version (1.86) to oldest multi-arch installer (1.65)

Tested:
  - Install
  - Removal
  - Installed name matches full_name

Not tested:
  - 32bit install/removal


samjhandley:
The cert does not cover the download.cpuid.com subdomain. 
Downloads are http only, is https a requirment for the state file?

twangboy:
@samjhandley We try to use https where we can. There was a PR a few weeks ago that went through and changed a bunch of HTTP urls to HTTPS.

If that's the case, and it seems like it is as I just tried it, please put a comment in there above the URL so that it doesn't get changed to HTTPS in the future.

samjhandley:
Added a comment :)

twangboy:
Thanks for the update!

PR #1530, opened 2019-01-26
Update

scheste:
Add current WinMerge version

### DIFF
--- a/winmerge.sls
+++ b/winmerge.sls
@@ -1,12 +1,15 @@
+{% set version = '2.16.0' %}
+
 {% if grains['cpuarch'] == 'AMD64' %}
     {% set PROGRAM_FILES = "%ProgramFiles(x86)%" %}
 {% else %}
     {% set PROGRAM_FILES = "%ProgramFiles%" %}
 {% endif %}
+
 winmerge:
-  '2.14.0':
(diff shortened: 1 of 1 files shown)


winmerge.sls
```diff
@@ -1,12 +1,15 @@
+{% set version = '2.16.0' %}
```
twangboy: Rather than remove old versions, we should make this a list of available versions.
```
(% set versions = ['2.16.0', '2.14.0']  %)
```

winmerge.sls
```diff
@@ -1,12 +1,15 @@
...
+
 {% if grains['cpuarch'] == 'AMD64' %}
     {% set PROGRAM_FILES = "%ProgramFiles(x86)%" %}
 {% else %}
     {% set PROGRAM_FILES = "%ProgramFiles%" %}
 {% endif %}
+
 winmerge:
```
twangboy: Add a for loop here after the package name (`winmerge`):
```
winmerge:
{% for version in versions %}
```

twangboy: Put an endfor at the bottom of the file to close the loop:
```
    ...
    reboot: False
{% endfor %}
```
See this for an example:
https://github.com/saltstack/salt-winrepo-ng/blob/master/pidgin.sls

PR #1532, opened 2019-01-29
update and rename state

scheste:
There are 2 different MSI packages, once per-user and once machine-wide

### DIFF
--- a/slack-user-msi.sls
+++ b/slack-user-msi.sls
@@ -1,5 +1,10 @@
-slack-msi:
-  '2.1.0':
+#MSI for per-user deployment package
+#The MSI for per-user deployment adds the Slack Installer machine-wide, rather than individually installing Slack for Windows. 
+#The Slack Installer will only add Slack for Windows if it was never installed for the user currently logged into the machine.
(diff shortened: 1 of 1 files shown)


slack-user-msi.sls
```diff
@@ -1,5 +1,10 @@
-slack-msi:
-  '2.1.0':
+#MSI for per-user deployment package
+#The MSI for per-user deployment adds the Slack Installer machine-wide, rather than individually installing Slack for Windows. 
+#The Slack Installer will only add Slack for Windows if it was never installed for the user currently logged into the machine.
+#https://get.slack.help/hc/en-us/articles/212475728-Deploy-Slack-via-Microsoft-Installer
+
+slack-user-msi:
+  '3.3.7.0':
```
twangboy: This looks like a situation where we will never really know the version of slack to be downloaded with the URL. It will always point to the latest version available. Chrome behaves in the same manner. In this case, instead of using a version number here, we just use `latest` as the version.

The drawback is that it will be reinstalled each time you do a highstate because there is no way to know if the version at the URL matches that installed on the system.

The only other option is to maintain your own copies of the MSI's on your system where you are tracking the version.

scheste: You're right, I'm changing the state and use latest instead of a version

PR #1538, opened 2019-02-01
Create vs_remotetools_2017

scheste:
Add initial version of Remote Tools for Visual Studio 2017

### DIFF
--- a/vs_remotetools_2017
+++ b/vs_remotetools_2017
@@ -1,0 +1,17 @@
+vs_remotetools_2017:
+  latest:
+    full_name: 'Remote Tools for Visual Studio 2017'
+    {% if grains['cpuarch'] == 'AMD64' %}
+    installer:   'https://aka.ms/vs/15/release/RemoteTools.amd64ret.enu.exe'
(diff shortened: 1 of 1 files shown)


vs_remotetools_2017
```diff
@@ -0,0 +1,17 @@
+vs_remotetools_2017:
+  latest:
```
twangboy: Does the Microsoft Issue updates to this software using the same URL? If so, then `latest` is appropriate here. However, the cost is that the software will be reinstalled each time you run the state. If it's a URL that is not changed often, or if you're not sure, I would recommend going with the actual version. `15.0.28307.280`

scheste: yes, I just added a new build version to our private winrepo a few weeks ago and now again. Unfortunately there is often not even a changelog to it. In these cases only the last 3 digits of the version have changed...
But I would have an idea how to improve the version specification `latest:`. 
Can I create the feature request in github or is it better via my support account?

twangboy: Do it in github. The windows package manager has been supported heavily by the open community. Best to get more eyes on the solution.

PR #1547, opened 2019-02-08
Update to latest version

scheste:
only the most recent version is included, because the download URL has changed slightly with the latest version


nunit-console.sls
```diff
@@ -1,9 +1,16 @@
+{% set versions = {'3.9':['0']} %}
```
twangboy: You need to find a way to support older versions of this software. You may need to create two loops. One that handles the `v` that is used in 3.9 and one that handles the rest where it's just the version without the `v`

scheste: I've added an extra loop for the path

PR #1549, opened 2019-02-08
Update to latest version

scheste:


adv-port-scanner.sls
```diff
@@ -1,13 +1,21 @@
...
-    installer: 'salt://win/repo-ng/adv-port-scanner/pscan24.exe'
+{% for major, subversions in versions.items() %}
+{% for minor in subversions %}
+  '{{ major }}.{{ minor }}':
+    full_name: 'Advanced Port Scanner {{ major }}'
+    installer: '{{ source_path }}/Advanced_Port_Scanner_{{ major }}.{{ minor }}.exe'
     install_flags: '/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP- & taskkill /F /T /IM advanced_port_scanner.exe & exit 0'
     uninstaller: 'msiexec.exe'
-    uninstall_flags: '/qn /noreboot /x {10F177CF-543F-4BC2-A297-DBF73709D3C5}'
+    uninstall_flags: '/qn /noreboot /x {76773275-7196-4B20-B3C6-27EDA624C053}'
```
twangboy: Is there any other way to uninstall this software without using the GUID? These can change between versions and would therefore not be a good default way to remove the software. Is there no `uninstaller` created?

PR #1550, opened 2019-02-08
Update to latest version

scheste:


bandizip.sls
```diff
@@ -1,7 +1,9 @@
+{% set source_path = 'http://ca-dl.bandisoft.com/bandizip' %}
+
 bandizip:
-  '5.12':
+  '6.20':
```
twangboy: We should probably keep the old version available as well. People who have this installed will no longer show as installed by salt if the version is just removed. Make it a list and add some jinja.

scheste: the problem is, that the download link 

The download link http://ca-dl.bandisoft.com/bandizip/BANDIZIP-SETUP.EXE always offers only the latest version for download. Maybe it would be even better in this case, instead of specifying a version number to use latest as version.

PR #1552, opened 2019-02-12
Add .sls file endings

stratusjerry:
@twangboy any thoughts on how the files in the root could be re-organized OR the travis test be written so that files not ending with .sls would fail?


twangboy:
Crap, I haven't been looking at the file names... just the code. I'll try to be more vigilant there. Let me ask @dwoz if he knows of a way to do that with Travis.

PR #1555, opened 2019-02-14
Initial Version Vivaldi Browser

scheste:


vivaldi.sls
```diff
@@ -0,0 +1,20 @@
+#Vivaldi is a Chromium-based browser
+#https://vivaldi.com
+
+{% set SOURCE_PATH = 'https://downloads.vivaldi.com/stable' %}
+
+vivaldi:
+  latest:
+    full_name: 'Vivaldi'
+    {% if grains['cpuarch'] == 'AMD64' %}
+    installer: '{{ SOURCE_PATH }}/Vivaldi.x64.exe'
```
twangboy: Looks like you can also download specific versions from the same location. For example...:
https://downloads.vivaldi.com/stable/Vivaldi.2.3.1440.41.x64.exe
Do you know if they maintain older versions there or just the latest? It would be interesting to watch this and see what happens on the next release. Will the link above still work or just the newer one.

It's better to use specific versions rather than the `latest` keyword if possible, to avoid the same software being repeatedly installed with every state run.

scheste: My first consideration was to use the version number, but Vivaldi behaves like Google Chrome. There are constantly newer build versions, often because of security vulnerabilities or minor bugfixes, which are offered as auto update. It is therefore not recommended to use an older version. There is also no public changelog for each build version so that older versions can be added to the state.

PR #1559, opened 2019-02-22
Create powershell-core

scheste:
Initial version of PowerShell Core


stratusjerry:
@scheste please make sure these are suffixed with .sls

twangboy:
@scheste I should have checked that as well

PR #1563, opened 2019-02-26
Update winrar

scheste:
urgent update due to vulnerabilities


winrar.sls
```diff
@@ -1,4 +1,5 @@
-{% set versions = {'5':['61']} %}
+#if possible use 5.70 or newer due to vulnerabilities (CVE-2018-20250, CVE-2018-20251, CVE-2018-20252, CVE-2018-20253)
+{% set versions = {'5':['70'], '5':['61']} %}
```
twangboy: I think the way this works is more along the lines of this:
```
{% set versions = {'5':['70','61']} %}
```

twangboy: You would only need to add another dictionary when the major version number changes

scheste: yes you're right

PR #1564, opened 2019-03-04
Create mariadb.sls

scheste:
Initial version of MariaDB


scheste:
I don't know why the check fails, I can download and install mariadb from my test environment with salt 2018.3.3.

scheste:
I have successfully installed and uninstalled all listed versions. Maybe it`s also due to the default mirror, that this only allows a certain number of downloads...

twangboy:
@TheBigBear Would you have any idea why these tests are failing?

scheste:
Changing the default mirror did not work for the Travis CI test, on my test system I can install and remove mariadb without any problems.

salt 'win2019' pkg.install mariadb
win2019:
    ----------
    mariadb:
        ----------
        new:
            10.3.13.0
        old:

salt 'win2019' pkg.remove mariadb
win2019:
    ----------
    mariadb:
        ----------
        new:
        old:
            10.3.13.0

twangboy:
@stratusjerry Any ideas why tests are failing here?

stratusjerry:
> @stratusjerry Any ideas why tests are failing here?

@twangboy Yes, it looks like that URL requires a mirrored repo to be appended to the end, which acts as a redirect to that mirror. Of note, not all mirrors have all packages. Based off the original file, I have created #1579 which uses a different URL that seems to work.

twangboy:
Closed in favor of https://github.com/saltstack/salt-winrepo-ng/pull/1579

PR #1575, opened 2019-03-29
Fixed vagrant sls

tomduijf:
The downloadable MSI installer filenames changed, now including arch.
Re-used the openvpn way of getting arch and translate it into either i686 or x86_64

Also updated the available versions list.


vagrant.sls
```diff
@@ -1,11 +1,11 @@
 vagrant:
-  {#{% for version in '1.8.6', '1.8.5',  '1.8.4', '1.8.1', '1.8.0', '1.7.4', '1.7.3', '1.7.2', '1.7.1', '1.6.5','1.6.4','1.6.3', '1.6.2', '1.6.1', '1.6.0', '1.5.4', '1.5.3', '1.5.2', '1.5.1', '1.5.0', '1.4.3', '1.4.2', '1.4.1', '1.4.0' %}#}
```
twangboy: Could you make two `for loops` here? One for the older version that are still in the old file format and then a second `for loop` for the new versions like your present code. We still need to support the older stuff.

Additionally, it makes it easier to update when the version numbers are in reverse numerical order... latest release up top.

tomduijf:
Reversed order
Added additional loop for older versions.

vagrant.sls
tomduijf: done

twangboy:
@tomduijf Perfect. Thanks!

PR #1576, opened 2019-03-30
Update virtualbox

scheste:
Latest version added and state improved

### DIFF
--- a/virtualbox.sls
+++ b/virtualbox.sls
@@ -1,118 +1,69 @@
+# Version 5.1 and older are no longer supported by Oracle
+# https://www.virtualbox.org/wiki/Download_Old_Builds
+
+{% set source_path = 'https://download.virtualbox.org/virtualbox/' %}
+
 virtualbox:
-  '5.2.12':
-    full_name: 'Oracle VM VirtualBox 5.2.12'
-    installer: 'https://download.virtualbox.org/virtualbox/5.2.12/VirtualBox-5.2.12-122591-Win.exe'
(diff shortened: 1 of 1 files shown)


virtualbox.sls
```diff
@@ -1,118 +1,69 @@
+# Version 5.1 and older are no longer supported by Oracle
+# https://www.virtualbox.org/wiki/Download_Old_Builds
+
+{% set source_path = 'https://download.virtualbox.org/virtualbox/' %}
+
 virtualbox:
-  '5.2.12':
-    full_name: 'Oracle VM VirtualBox 5.2.12'
-    installer: 'https://download.virtualbox.org/virtualbox/5.2.12/VirtualBox-5.2.12-122591-Win.exe'
+{% for version, build_number in [('6.0.4', '128413')] %}
```
twangboy: These can be combined in to a list of sets:
```
{% set versions = [('6.0.4', '128413'),
                   ('5.2.26', '128414'),
                   ('5.1.38', '122592')] %}
{% for version, build_number in versions %)
```
That should eliminate most of the repetitive code.

scheste: I've already thought about that, but then I'd have to add some if loops because of the different GUIDs (every version + different for 32 and 64 bit). Unfortunately Virtualbox can still only be uninstalled using the GUID.
Therefore undecided whether it would not become too complex and whether it would not be simpler?

twangboy: Damn...

PR #1577, opened 2019-03-30
Update dropbox

scheste:
updated to the latest version

### DIFF
--- a/dropbox.sls
+++ b/dropbox.sls
@@ -1,16 +1,21 @@
 # just 32-bit x86 installer available
+
+{% set SOURCE_PATH = 'https://clientupdates.dropboxstatic.com/dbx-releng/client/' %}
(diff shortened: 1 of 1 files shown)


dropbox.sls
```diff
@@ -1,16 +1,21 @@
 # just 32-bit x86 installer available
+
+{% set SOURCE_PATH = 'https://clientupdates.dropboxstatic.com/dbx-releng/client/' %}
+{% set VERSION = '69.4.102' %}
```
twangboy: Please keep the old version available as well.

scheste: I don't know if it makes sense to add an older versions, because dropbox also has an auto-update function and older clients are updated automatically anyway

twangboy: OK

PR #1583, opened 2019-04-25
Updating NXLog latest version to 2.10.2150

couilllard45682:


nxlog.sls
```diff
@@ -1,5 +1,5 @@
 nxlog:
-  {% for version in ['2.10.2102',
+  {% for version in ['2.10.2150', '2.10.2102',
```
twangboy: The older versions are causing a 404 error. So, unless you can find the URL to the older installer files, we'll probably need to remove the older ones. If that's the case, there's no point in having all the jinja code. But, I'm fine if you leave it in.

couilllard45682:
Updated to remove older and unavailable versions

nxlog.sls
couilllard45682: Done

nxlog.sls
```diff
@@ -1,7 +1,5 @@
 nxlog:
-  {% for version in ['2.10.2102',
-                     '2.9.1716'] %}
-  '{{ version }}':
+  '2.10.2150':
     full_name:  'NXLog-CE'
     installer: 'https://nxlog.co/system/files/products/files/348/nxlog-ce-{{ version }}.msi'
```
twangboy: Need to remove `{{ version }}` here and hard-code the version

twangboy: Same for the uninstaller

PR #1584, opened 2019-04-25
Update node.js-lts.sls

cncshaggy:
Updating to the 3 most recent LTS branches

I used [this link](https://nodejs.org/en/download/releases/) to get the versions


node.js-lts.sls
```diff
@@ -1,5 +1,5 @@
 node.js-lts:
-  {% for version in ['6.9.1','6.11.3'] %}
+  {% for version in ['10.15.3', '8.16.0', '6.17.1'] %}
   '{{ version }}':
```
twangboy: Please leave older versions and just add the new.

cncshaggy:
Adding the old versions back

PR #1585, opened 2019-04-25
Update node.js.sls

cncshaggy:
Updating to the 3 most recent LTS branches

I used [this link](https://nodejs.org/dist/latest-v11.x/) to get the version

### DIFF
--- a/node.js.sls
+++ b/node.js.sls
@@ -1,5 +1,5 @@
 node.js:
-  {% for version in ['7.0.0'] %}
+  {% for version in ['12.0.0', '11.14.0'] %}
   '{{ version }}':
     full_name: 'Node.js'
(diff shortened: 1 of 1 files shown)


node.js.sls
```diff
@@ -1,5 +1,5 @@
 node.js:
-  {% for version in ['7.0.0'] %}
+  {% for version in ['12.0.0', '11.14.0'] %}
```
twangboy: Please leave 7.0 for those who have it installed. Just add the newer versions. All versions still available here: https://nodejs.org/dist/

cncshaggy:
Update with the version array `['12.0.0', '11.14.0', '10.15.3', '9.11.2', '8.16.0', '7.10.1', '7.0.0']`

PR #1594, opened 2019-05-12
Update npp.sls

TheBigBear:


TheBigBear:
@twangboy is this 'my' travis CI or your 'saltstack' Travis CI that is failing? If it is yours, can you see why they fail?

twangboy:
@TheBigBear I've never set up TravisCI for this. I'm not sure who set it up. I should probably know though.

PR #1597, opened 2019-05-12
Update pdf24creator.sls

TheBigBear:


pdf24creator.sls
```diff
@@ -1,5 +1,5 @@
 pdf24creator:
-  {% for version in ['8.5.0'] %}
+  {% for version in ['8.8.2'] %}
```
twangboy: Looks like 8.5.0 is still available. Can we just add the newest version to the list? Or maybe add all the ones in between...
```
['8.8.2', '8.8.1', '8.8.0', '8.7.2', '8.7.1', '8.7.0', '8.6.1', '8.6.0', '8,5.0']
```

TheBigBear: @twangboy OK, done that

PR #1606, opened 2019-05-18
Add CCleaner slim edition start to 5.57 version

kartnico:


TheBigBear:
@kartnico is the non slim version no longer downloadable for some reason?

kartnico:
> 
> 
> @kartnico is the non slim version no longer downloadable for some reason?

No it is always available. But I add the possibility to install also the slim edition.

So in parallel, I also update the full CCleaner edition, and I add cclean slim edition.

PR #1609, opened 2019-05-21
Update NTP to 4.2.8p13

nickj:
Notably, CVE-2018-12327 fixed in 4.2.8p12

### DIFF
--- a/ntp.sls
+++ b/ntp.sls
@@ -9,7 +9,7 @@
     {% set PROGRAM_FILES = "%ProgramFiles%" %}
 {% endif %}
 ntp:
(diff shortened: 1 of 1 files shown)


ntp.sls
```diff
@@ -9,7 +9,7 @@
     {% set PROGRAM_FILES = "%ProgramFiles%" %}
 {% endif %}
 ntp:
-  {% for version in ['4.2.8p11'] %}
+  {% for version in ['4.2.8p11', '4.2.8p12', '4.2.8p13'] %}
```
twangboy: Can you reverse the order here... newer versions first. Makes it easier to add when you have 100 versions.

nickj: Sure, done now!

PR #1618, opened 2019-06-19
Update chocolatey package to always install the latest version

kartnico:


TheBigBear:
@kartnico thanks for this. I am not sure it makes sense to use the version number in this way? they never have more than one version available online. would it make sense to put the version number into the 'cmd' file you downloaded at the time and put that in your local repo? I would vote for either only ever listing one version number as only one is available online, or else actually use the version number in a renamed install.cmd script off the local repo?

twangboy:
@kartnico @TheBigBear If that's the case, we should treat it in the same way we do the Software Definition Package for Chrome.... where there's only ever one binary available for Chrome that always installs the latest.

kartnico:
@TheBigBear and @twangboy is it ok for you ?

PR #1620, opened 2019-06-20
Update azuredatastudio.sls

apoloniogg:
This change allows the installer to be downloaded from the https source.


TheBigBear:
@apoloniogg thanks for  your submission, have you tested this? There is no version number provided, so I don't think this has any chance to work?

PR #1622, opened 2019-06-21
Zabbix Agent installer

absmith82:
Currently this only works for x64 based windows as I don't have a x86 based windows server and the agent won't allow you to do an x86 based install on the x64 based server.

Also I have been unable to get the agent to uninstall via cli in quiet mode properly, windows tells me the client is not installed or other things and so the uninstall fails unless you do a full removal from the windows programs and features center.

### DIFF
--- a/zabbix-agent.sls
+++ b/zabbix-agent.sls
@@ -5,6 +5,8 @@ zabbix-agent:
     installer: 'https://www.zabbix.com/downloads/4.2.3/zabbix_agent-4.2.3-win-amd64-openssl.msi'
     {% endif %}
     install_flags: '/quiet SERVER=localhost'
+    uninstaller: 'msiexec.exe'
+    uninstall_flags: '/x https://www.zabbix.com/downloads/4.2.3/zabbix_agent-4.2.3-win-amd64-openssl.msi /quiet'
     locale: en_US
     reboot: False
   '4.0.9.2400':
@@ -13,6 +15,8 @@ zabbix-agent:
     installer: 'https://www.zabbix.com/downloads/4.0.9/zabbix_agent-4.0.9-win-amd64-openssl.msi'
     {% endif %}
     install_flags: '/quiet SERVER=localhost'
+    uninstaller: 'msiexec.exe'
+    uninstall_flags: '/x https://www.zabbix.com/downloads/4.0.9/zabbix_agent-4.0.9-win-amd64-openssl.msi /quiet'
     locale: en_US
     reboot: False
   '3.0.28.2400':
@@ -21,5 +25,7 @@ zabbix-agent:
     installer: 'https://www.zabbix.com/downloads/3.0.28/zabbix_agent-3.0.28-win-amd64-openssl.msi'
     {% endif %}
     install_flags: '/quiet SERVER=localhost'
+    uninstaller: 'msiexec.exe'
+    uninstall_flags: '/x https://www.zabbix.com/downloads/3.0.28/zabbix_agent-3.0.28-win-amd64-openssl.msi /quiet'
     locale: en_US
     reboot: False
--- a/zabbix-agent.sls
+++ b/zabbix-agent.sls
@@ -6,16 +6,18 @@ zabbix-agent:
 {% for major, subversions in versions.items() %}
 {% for minor in subversions %}
   '{{major}}.{{minor}}.2400':
-    full_name: 'Zabbix Agent (64-bit)'
     {% if grains['cpuarch'] == 'AMD64' %}
+    full_name: 'Zabbix Agent (64-bit)'
     installer: '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-amd64-openssl.msi'
-    uninstall_flags: '/x {{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-amd64-openssl.msi /quiet'
+    uninstaller: '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-amd64-openssl.msi'
     {% else %}
+    full_name: 'Zabbix Agent'
     installer: '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-i386-openssl.msi'
-    uninstall_flags: '/x {{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-i386-openssl.msi /quiet'
+    uninstaller:  '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-i386-openssl.msi'
     {% endif %}
-    install_flags: '/quiet SERVER=localhost'
-    uninstaller: 'msiexec.exe'
+    install_flags: '/qn SERVER=localhost'
+    uninstall_flags: '/qn /norestart'
+    msiexec: True
     locale: en_US
     reboot: False
 {% endfor %}
--- a/zabbix-agent.sls
+++ b/zabbix-agent.sls
@@ -1,5 +1,5 @@
 zabbix-agent:
-  4.2.3.2400:
+  '4.2.3.2400':
     {% if grains['cpuarch'] == 'AMD64' %}
     full_name: 'Zabbix Agent (64-bit)'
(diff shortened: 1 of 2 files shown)


TheBigBear:
@absmith82 you need to enclose the version numbers in single quotes. python yaml misinterpret version numbers that aren't quoted. doing that might even fix your inconsistent behaviour you have seen, as without it salt sends the wrong instructions ...

absmith82:
added the single quotes,  the functionality works the same with them or without them so far.  The uninstaller issue doesn't seem to be a salt issue as much as an issue with the MSI package.  I can't get it to do a quite uninstall on the system via cli at all through salt or not.  It will do a regular uninstall but that sort of defeats the purpose of having the uninstaller in salt.

absmith82:
I found a way to fix the uninstaller,  I had to call the msi package directly from the website to get it to work, but it is working now.

zabbix-agent.sls
```diff
@@ -0,0 +1,31 @@
+zabbix-agent:
+  '4.2.3.2400':
+    {% if grains['cpuarch'] == 'AMD64' %}
```
scheste: The IF loop should be removed if you don't have a 32Bit installer or higher-level that if the system is not a 64Bit system, the entire state is bypassed

zabbix-agent.sls
```diff
@@ -0,0 +1,31 @@
+zabbix-agent:
+  '4.2.3.2400':
```
scheste: if you want to make the state more beautiful, you can create multiple variable.
e.g. for the version or installer path, you can look at other states (7zip.sls as an example)

absmith82:
I can definitely remove the if loop.  I put it in there because there is a
32-bit installer that I haven't been able to test because I don't have a
32-bit image to work with at the moment and was going to put it in once I
do get that up and running and tested.  The 64-bit installer obviously will
not work on a 32-bit system and the 32-bit installer is designed not to
allow installation on a 64-bit system.

I will see if I can make the state function more like the 7zip state on
iterations of the versions rather than having to write a new portion of the
state each time, That actually would be much cleaner and easier to
maintain.  Thanks.

*Adam Smith*
Senior Systems Engineer

Sundance Institute
O:435.658.3456
E:adam_smith@sundance.org
www.sundance.org

<http://www.sundance.org>

absmith82:
Thanks for pointing me at the 7zip installer, I have now modeled the versioning with variables like in that state and have also added the 32-bit installer.

zabbix-agent.sls
```diff
@@ -0,0 +1,22 @@
...
+  '{{major}}.{{minor}}.2400':
+    full_name: 'Zabbix Agent (64-bit)'
+    {% if grains['cpuarch'] == 'AMD64' %}
+    installer: '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-amd64-openssl.msi'
+    uninstall_flags: '/x {{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-amd64-openssl.msi /quiet'
+    {% else %}
+    installer: '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-i386-openssl.msi'
+    uninstall_flags: '/x {{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-i386-openssl.msi /quiet'
+    {% endif %}
+    install_flags: '/quiet SERVER=localhost'
```
scheste: you should use **/qn** instead of **/quiet** parameter, otherwise it can happen that the installation fails

zabbix-agent.sls
```diff
@@ -0,0 +1,22 @@
...
+    uninstaller: 'msiexec.exe'
```
scheste: You should use the same path as for the `installer:` (32 + 64Bit) parameter and add the following parameter:
`    msiexec: True`
Because if you have an MSI package, you can use this parameter

zabbix-agent.sls
```diff
@@ -0,0 +1,22 @@
...
+{% for major, subversions in versions.items() %}
+{% for minor in subversions %}
+  '{{major}}.{{minor}}.2400':
+    full_name: 'Zabbix Agent (64-bit)'
+    {% if grains['cpuarch'] == 'AMD64' %}
+    installer: '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-amd64-openssl.msi'
+    uninstall_flags: '/x {{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-amd64-openssl.msi /quiet'
+    {% else %}
+    installer: '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-i386-openssl.msi'
+    uninstall_flags: '/x {{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-i386-openssl.msi /quiet'
```
scheste: remove the installer from the `uninstall_flags` and use only the parameter `/qn /norestart`

absmith82:
That doesn't actually seem to work for this package.  Having the installer
in the flag was the only way it would properly uninstall without
interaction. Even via cli.  /qn /norestart is fine and probably safer but
the only way it would uninstall for me with no interaction was to include
the installer in the flags.  Even using the installer as the uninstall
method instead of msiexec didn't seem to work properly realizing that all
that would be doing is calling msiexec because of the msi package.  Or
using msiexec with the program uuid still didn't work for me.

zabbix-agent.sls
absmith82: That doesn't actually seem to work for this package.  Having the installer
in the flag was the only way it would properly uninstall without
interaction. Even via cli.  /qn /norestart is fine and probably safer but
the only way it would uninstall for me with no interaction was to include
the installer in the flags.  Even using the installer as the uninstall
method instead of msiexec didn't seem to work properly realizing that all
that would be doing is calling msiexec because of the msi package.  Or
using msiexec with the program uuid still didn't work for me.

absmith82: Was finally able to make this work with the msiexec: True flag for the uninstaller.  I must have been including the /x flag for uninstall_flags on the state when it wasn't needed.

absmith82:
Anything blocking this from moving forward at this point?

zabbix-agent.sls
```diff
@@ -0,0 +1,24 @@
...
+    {% if grains['cpuarch'] == 'AMD64' %}
+    full_name: 'Zabbix Agent (64-bit)'
+    installer: '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-amd64-openssl.msi'
+    uninstaller: '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-amd64-openssl.msi'
+    {% else %}
+    full_name: 'Zabbix Agent'
+    installer: '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-i386-openssl.msi'
+    uninstaller:  '{{source_path}}{{major}}.{{minor}}/zabbix_agent-{{major}}.{{minor}}-win-i386-openssl.msi'
+    {% endif %}
+    install_flags: '/qn SERVER=localhost'
```
twangboy: Please add `/norestart` to the install_flags as well, please.

absmith82: Fixed that issue

absmith82:
Install flags have now been fixed.  Let me know if there is anything else.

zabbix-agent.sls
```diff
@@ -0,0 +1,24 @@
+# both 32-bit (x86) AND a 64-bit (AMD64) installer available
+{% set versions = {'4.2':['3'], '4.0':[9], '3.0':[28]} %}
+{% set source_path = 'https://www.zabbix.com/downloads/' %}
+
+zabbix-agent:
+{% for major, subversions in versions.items() %}
+{% for minor in subversions %}
+  '{{major}}.{{minor}}.2400':
```
twangboy: Sorry, just one more question. Do all versions have a build number of `2400`?

twangboy: Just installed a few... yep... 2400

twangboy:
@absmith82 Thanks for the PR!

absmith82:
Thanks for making this much easier on me for zabbix installs.

PR #1636, opened 2019-08-29
Update vlc.sls

stratusjerry:
Add new version fixing CVEs and remove unsupported version

### DIFF
--- a/vlc.sls
+++ b/vlc.sls
@@ -1,7 +1,7 @@
 # both 32-bit (x86) AND a 64-bit (AMD64) installer available
 {% set PROGRAM_FILES = "%ProgramFiles%" %}
 vlc:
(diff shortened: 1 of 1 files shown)


vlc.sls
```diff
@@ -1,7 +1,7 @@
 # both 32-bit (x86) AND a 64-bit (AMD64) installer available
 {% set PROGRAM_FILES = "%ProgramFiles%" %}
 vlc:
-{% for version in ['3.0.7.1', '3.0.7', '3.0.6', '3.0.5', '3.0.4', '3.0.3', '3.0.2', '3.0.1', '3.0.0', '2.2.8', '2.2.7', '2.2.6'] %}
+{% for version in ['3.0.8', '3.0.7.1', '3.0.7', '3.0.6', '3.0.5', '3.0.4', '3.0.3', '3.0.2', '3.0.1', '3.0.0', '2.2.8', '2.2.6'] %}
```
twangboy: Is there a reason why you removed 2.2.7? Is that no longer available?

stratusjerry: Correct, seems it was a macos only fix release. There is no Windows installer available for that version

PR #1644, opened 2019-10-16
add nextcloud client

colttt:


colttt:
also add latest libreoffice business version

nextcloud-client.sls
```diff
@@ -0,0 +1,11 @@
+nextcloud-client:
```
twangboy: Please put more recent versions at the top... to the front.

PR #1665, opened 2020-01-23
feat(zabbix-agent): Update available versions and msi url paths

dth202:
Added available zabbix versions since last update.


absmith82:
Just wanted to note that there is now 3.0.29 as well and was wondering if you would be able to add 29 to the minor versions of 3.0  

Thanks, I'm glad someone else is using this package and keeping it up to date as well!!!

PR #1673, opened 2020-02-26
Keepass 2x version logic patch and add versions 2.39 to 2.44

TerraTalpi:
I have changed the logic of the keepass-2x versions from just every single instruction to an array-based version list. All of the installers are msi packages and therefore installed equally. I have tested the installation and correct identification of the versions as well.

Finally I have added the version 2.39 to 2.44.


keepass-2x.sls
```diff
@@ -1,109 +1,26 @@
...
+    uninstaller: 'https://downloads.sourceforge.net/project/keepass/KeePass%202.x/{{ version }}/KeePass-{{ version }}.msi'
     uninstall_flags: '/qn /norestart'
     msiexec: True
     locale: en_US
     reboot: False
-  '2.39.1':
-    full_name: 'KeePass 2.39.1'
-    installer: 'https://downloads.sourceforge.net/project/keepass/KeePass%202.x/2.39.1/KeePass-2.39.1.msi'
+{% endfor %}
+{% for version in ['2.44', '2.43', '2.42', '2.41', '2.40', '2.39', '2.38', '2.37', '2.36', '2.35', '2.34', '2.33', '2.32', '2.31', '2.30', '2.29'] %}
```
twangboy: Could you add `.0` to all these and then include the 2 versions from above so that it's all sharing the same code? That is, if that's the only difference.

TerraTalpi: Hello @twangboy, I had the same intention but there are two problems with that:
1) the saltstack "fullname" field has to exactly match the installed software name. Keepass installs itself with the fullname `KeePas 2.44`. If I change the `{{ version }}` to e.g. `2.44.0` it would not match anymore and the software would not be recognized by saltstack anymore.
1) the download URL also requires `2.44` rather than `2.44.0`

The `.0` is only used for the saltstack version, Keepass always omits it.

The only solution I can think of here is a conditional statement... if the `{{ version }}` has only one dot, then add `.0` to it, otherwise take it as it its. Would that be better?

twangboy: Cool. Thanks for the explanation.

TerraTalpi:
Ok, from my side ready to merge. Please review once more and if applicable merge.

PR #1724, opened 2020-10-09
Add support for new prometheus windows_exporter versions

maximecharron:
One notable change is that starting with version 0.13.0, wmi_exporter is now named windows_exporter.

The repo has also moved from  https://github.com/martinlindhe/wmi_exporter to https://github.com/prometheus-community/windows_exporter. All the releases were also moved so it's safe to update the URL for every version.


maximecharron:
@twangboy Since I see you recently approved some reviews, do you think you could take a look at this? Thanks! :)

twangboy (approved):

PR #1736, opened 2020-10-25
Update npp.sls to version 7.9

TheBigBear:
rewritten all 7.x series available downloads as they moved the downloads to GitHub


TheBigBear:
@twangboy or @dafyddj can you please take a look at the recently changed travisci output? I do not see any proof that it actually ran the URL tests - and succeeded? If memory serves me well then it used to show some lines for each URK. Could it be that the new pre-commit travis parts have an issue ? If I read it right it checked tens of URLs in 0.24s ? I suspect it has an exception and did not actually check the URLs? @twangboy 
```
Check URLs exist and are valid...........................................Passed
276- hook id: check-urls
277- duration: 0.24s
278
279---( npp.sls )---
280[EXCEPTION] <class 'jinja2.exceptions.TemplateSyntaxError'>
281Traceback (most recent call last):
282  File ".travis/tests.py", line 163, in <module>
283    t = Template(template)
284  File "/home/travis/.cache/pre-commit/repo1a66503c/py_env-python3/lib/python3.6/site-packages/jinja2/environment.py", line 1031, in __new__
285    return env.from_string(source, template_class=cls)
286  File "/home/travis/.cache/pre-commit/repo1a66503c/py_env-python3/lib/python3.6/site-packages/jinja2/environment.py", line 941, in from_string
287    return cls.from_code(self, self.compile(source), globals, None)
288  File "/home/travis/.cache/pre-commit/repo1a66503c/py_env-python3/lib/python3.6/site-packages/jinja2/environment.py", line 638, in compile
289    self.handle_exception(source=source_hint)
290  File "/home/travis/.cache/pre-commit/repo1a66503c/py_env-python3/lib/python3.6/site-packages/jinja2/environment.py", line 832, in handle_exception
291    reraise(*rewrite_traceback_stack(source=source))
292  File "/home/travis/.cache/pre-commit/repo1a66503c/py_env-python3/lib/python3.6/site-packages/jinja2/_compat.py", line 28, in reraise
293    raise value.with_traceback(tb)
294  File "<unknown>", line 8, in template
295jinja2.exceptions.TemplateSyntaxError: expected token ',', got 'float'
296--------------------------------------------------------------------------------
297Everything went smoothly. No errors were found. Happy deployment!
```

TheBigBear:
Sorry folks, I just spotted it was a typo in the jijnja for version, will fix that, but the travis tests surely should have picked that up? 
```
had '7.6.1', ''7.6, '7.5.9',
                 ^^^^^^
```

TheBigBear:
@dafyddj sorry, I submitted two more commits to my PR and thereby removed the errors that I talked about above. So it will be hard for you to follow what was 'wrong', unless travis has more history? 
First issue was that, my Jinja in version line had a typo and even though 'something' saw that it encountered a float when expecting an INT it did not flag it as a Jinja error, but said all was fine.
The second non-issue was, that notepad plus plus no longer has all versions of the installer online on GitHub. So had to come up with a 'new' way of allowing uninstalls of old versions without allowing installs using installers that are no longer online. For that I chose to copy what salt-minion does for removing CVE versions from being installable. Does that seem like a good idea ( and is it necessary ? ) As the pkg full name does not reflect the version number in any way this is probably superfluous in the  case of notepad plus plus and we could simply drop off every version that no longer has an installer available , right? Or am I missing something here? @dafyddj @twangboy

dafyddj:
> Sorry folks, I just spotted it was a typo in the jijnja for version, will fix that, but the travis tests surely should have picked that up?
> 
> ```
> had '7.6.1', ''7.6, '7.5.9',
>              ^^^^^^
> ```

You're right, the Jinja error should have been picked up as a failure - I thought we had corrected that problem recently but I'll look into it further.

twangboy (approved):

PR #1766, opened 2020-11-24
Docker Desktop Package

davewoodward:
Added sls files to support installation of docker-desktop on Windows 10:

- wsl2-update.sls (Windows Subsystem for Linux Update is Prerequisite for Docker Desktop)

- docker-desktop.sls (Docker Desktop)


davewoodward:
The check is failing because the Docker Desktop endpoint is not including a content type response header.  I'm wondering if it would be acceptable to remove this from the tests since it doesn't look like the HTTP standard requires it.  I understand the desire to do everything possible to ensure that the URL specified returns a valid installer but it doesn't look like we can guarantee that the content type header will always be returned.

davewoodward:
I went ahead and removed the content type check from the test.  Let me know if this is unacceptable, and what alternative you recommend, and I can revert the changes and implement the alternative.

dafyddj:
Hi,
this seems like throwing out the baby with the bathwater - and I would say it renders the testing useless for all packages.

The alternative is to download the file and check it's an executable - but I wonder whether it's possible to just download a small portion of the file (100-200k) and check that it's a binary file or at least not HTML.

davewoodward:
I can do a little digging to see if there is better check that can be done to verify that the file is binary.

davewoodward:
@dafyddj  - I added a mime type check of the URL as a fallback if the content type header is not returned.  Seemed like a decent compromise.  I'm actually questioning how much (if any) additional value there is in the content type check since the mime type check will return "None" if the URL is invalid or the mime type can't be determined.

davewoodward:
Let me know if you have any other feedback or need anything from me to push this forward.

davewoodward:
@twangboy - any chance of getting this merged in the near future?  If not, then I'll have to make other arrangements to support the required functionality.  I know you haven't been assigned to this but you seem to take a lot of ownership over this repo so I was hoping you could get me some feedback.

dafyddj:
> @dafyddj - I added a mime type check of the URL as a fallback if the content type header is not returned. Seemed like a decent compromise. I'm actually questioning how much (if any) additional value there is in the content type check since the mime type check will return "None" if the URL is invalid or the mime type can't be determined.

I'm afraid this test does nothing to check whether an installer exists or not.

I have a fix for this coming but it will probably be a few days to get it merged.
In the meantime can I suggest you revert your test changes.

I notice that the installers specified in your sls files are not tied to a particular version (and there is no archive of past installers). In this case the version "latest"  is more appropriate I think (see the Chrome package for example).

In case the delay here is affecting your work, I wonder whether you're aware that it's pretty straightforward to keep a small repo of your own packages with your states. You just need the sls files to appear in the Salt file hierarchy under `salt://win/repo-ng`. I do this for a dozen or so packages that are not appropriate for a public repo.

davewoodward:
> > @dafyddj - I added a mime type check of the URL as a fallback if the content type header is not returned. Seemed like a decent compromise. I'm actually questioning how much (if any) additional value there is in the content type check since the mime type check will return "None" if the URL is invalid or the mime type can't be determined.
> 
> I'm afraid this test does nothing to check whether an installer exists or not.
> 
> I have a fix for this coming but it will probably be a few days to get it merged.
> In the meantime can I suggest you revert your test changes.
> 
> I notice that the installers specified in your sls files are not tied to a particular version (and there is no archive of past installers). In this case the version "latest" is more appropriate I think (see the Chrome package for example).
> 
> In case the delay here is affecting your work, I wonder whether you're aware that it's pretty straightforward to keep a small repo of your own packages with your states. You just need the sls files to appear in the Salt file hierarchy under `salt://win/repo-ng`. I do this for a dozen or so packages that are not appropriate for a public repo.

I'm confused at your comment that this does not test whether the installer exists or not.  The process is still using curl to make a HEAD request to the hosting HTTP server and is validating the return code.  The only change that was made is in checking the mime type of the response in cases where the HEAD response didn't include a content type header.  In this exceptional condition, we will know that the file existed because the HTTP HEAD request returned a 200 and we will know that the file extension identifies a binary file type.

I chose to leave the Docker Desktop install version-specific (instead of latest) to support future versions.  I can go ahead and go back a few versions but didn't know how far back you normally go with new Windows packages submitted to this repo.  I hate to simply go with latest because that prevents DevOps engineers from locking down the version.  Let me know if there are any standards on how far back to go on supporting old versions with a new package like this.  I can go back all the way if needed but that may not provide much value and we may get into some installer inconsistencies.

twangboy:
@davewoodward Looks like we have some merge conflicts. Could you please resolve?

davewoodward:
> @davewoodward Looks like we have some merge conflicts. Could you please resolve?

@twangboy - Done

twangboy (approved):

PR #1769, opened 2020-12-15
Test improvements: more accurate metrics / use python-magic where no content-type is returned

dafyddj:
Fixes 63 checks. Introduces 4 new failures (1 pkg) due to the HTTP polling method.


twangboy (approved):

dafyddj:
Thanks!

PR #1773, opened 2020-12-24
Add more Zabbix versions and fix download path

timwsuqld:


twangboy (approved):

twangboy:
@dafyddj Would you have any idea why these urls are failing? They seem valid.

timwsuqld:
@twangboy After some investigations, it seems the same versions are not available for Windows and Linux.
e.g. Linux select box at https://www.zabbix.com/download_agents?version=4.0+LTS&release=4.0.27&os=Linux&os_version=3.0&hardware=amd64&encryption=No+encryption&packaging=Archive has 4.0.15 -> 4.0.27. While the Windows 4.0 LTS OpenSSL has 4.0.6, 4.0.7, 4.0.9 -> 4.0.17, 4.0.22->4.0.27. Windows 4.0 LTS without OpenSSL has 4.0.0, 4.0.4 -> 4.0.7, 4.0.9 -> 4.0.27.

I'll try and update it for the correct Windows OpenSSL versions

timwsuqld:
@twangboy @dafyddj After looking through the travis code and the build log, I saw it was a 403 error. I belive it's a user agent issue, not sure why the zabbix CDN is blocking us. I got it to work by adding the following at line 107 of `tests.py`
```python
                c.setopt(curl.USERAGENT, 'Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; Googlebot/2.1; +http://www.google.com/bot.html) Safari/537.36')
```

Not sure what the correct useragent should be, and we might want to talk to whomever runs the Zabbix CDN to find out why it was getting blocked.

twangboy:
@dafyddj What do you think about adding that user agent?
@timwsuqld I tried updating this PR with your changes, but you aren't allowing maintainers to modify this PR.

timwsuqld:
@twangboy I'm not sure why but Github won't let me change the PR to allow maintainers to update it. After some testing, any useragent that has PycURL in it will be blocked, error 1010 (https://support.cloudflare.com/hc/en-us/articles/360029779472-Troubleshooting-Cloudflare-1XXX-errors#error1010). Before I update the PR with a different user-agent I'm going to reach out to the Zabbix team to work out the reason behind this. Maybe we can come up with our own user-agent that won't be blocked, once they understand we are doing HEAD only requests to check that the files exist.

timwsuqld:
@twangboy Zabbix have fixed the CDN to not block the CI tests. I can't see any way to retrigger Travis but it works for me locally when I run `tests.py` now.

twangboy:
Looks like we can't do any builds right now:
```
Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.
```

twangboy:
@timwsuqld Please rebase this. Then we should have some running tests.

timwsuqld:
@twangboy Rebased and CI has worked now.

PR #1789, opened 2021-03-26
Update all npp versions to latest incl intermediates

TheBigBear:


dafyddj:
Possibly a mistake but you've added winscp to this PR.

TheBigBear:
@dafyddj yes, not intentional. I am working on a number of updates. but the way the PR work is that if the first patch has not been merged yet, then the diff is the diff of the first and second ... etc etc 
How do I stop this from happening, so that they are 'clean' and only contain one programs updates in each , but I am still able to submit the three or four patches I am working on ?

TheBigBear:
@dafyddj do you know how I can remove the winscp commits from this PR ?

dafyddj:
From what I can see you are creating your PRs from your own `master` branch, so each time you push to that branch it updates any PR created from that branch.
You should rather create your PRs from a feature branch such as `update-npp` for example. More [here](https://reflectoring.io/github-fork-and-pull/#working-on-multiple-pull-requests-at-once).

dafyddj:
> @dafyddj do you know how I can remove the winscp commits from this PR ?

This is a bit more advanced but you would need to "reset" your master branch to just after the `npp` commit.
Something like `git reset --hard d40772c`, but be careful because this is destructive and you have to know what you're doing with it.

TheBigBear:
@dafyddj thanks, I will do them from a feature branch in future. ( I think I used to do that in the past - but it has been many months since my last patches - and I forgot all about that. ) I will delete / close the npp PR and re-create it new from a feature branch. probably safest and fastest at this stage. Thanks for your help.

PR #1790, opened 2021-03-26
Update RabbitMQ

keslerm:
Added all RabbitMQ versions for the current minor release (3.8)


TheBigBear:
@keslerm hi I would suggest that you  should not lump different programs into one combined PR , if there is any push back or corrections needed for one, it will be denied, best to submit them one PR by program I think.

TheBigBear:
The reason they usually want all releases, is that salt can then match on whatever version is present and do the relevant upgrades or downgrades. And in the case where the older versions are no longer publicly downloadable they still want the info on those older versions kept even if only to allow clean uninstall or removals of those versions. ( see npp.sls for one example of the format they prefer for that scenario )

keslerm:
@TheBigBear I split them up and added versions back to the current minor release (3.8)

I could add back to 3.7 but 3.8.0 came out in 2019 i feel like it's probably safe to stop there but I can keep going if it seems useful.

twangboy (approved):

PR #1791, opened 2021-03-26
WIP: Update Erlang

keslerm:
Added all Erlang versions for the last two major releases (23.x and 22.x)

Edit: Need to test this a bit more because erlang versioning is all over the place


keslerm:
Erlang has a weird caveat where minor versions ending in 0 are truncated to just show the major version.

So 23.0 shows up as 23 where as 23.3 shows up at 23.3

I attempted to deal with it the best way i could think of but i'm open to suggestions

twangboy (approved):

PR #1801, opened 2021-03-28
Update atom.sls

TheBigBear:


TheBigBear:
@twangboy have you guys in saltstack/vmware had any discussions at all about the status of the winrepo-ng and it's future? Who chooses what windows pkgs to maintain? And at what frequency, etc ? Any plans to replace it with something a bit more modern and more easily sustained ( and with a bit more security ) as this system is just so open for abuse and mis-use from a security point of view. ( no hashes, no checks, no digital signatures, no approved caches ... all things which have been becoming a must for software pkg managers in recent years IMHO )

twangboy:
@TheBigBear It's good to see you. It's been a while. We haven't had any discussions with vmware about winrepo-ng. Though I agree with you regarding the security features that could be added. The only discussion we have had on the core team is creating a module and state for the winget project (https://github.com/microsoft/winget-cli). The problem is that it only supports Windows 10. Not even sure if it supports Server versions of Windows.

twangboy (approved):

TheBigBear:
@twangboy winget is very  much tied to the windows store right now anyway and server os support is not ( yet ) not even on their roadmap. There are strong opinions out there of winget not belonging on server os , but I guess time will tell. But at the moment it looks server os support would only possibly come after they reached 1.0 and decide to add it to their roadmap in the future. Thanks for all the merges you di yesterday. Who else has merge rights and can be pinged to pick up or keep going, so as not to waste too much of your precious time for such a menial task as merging winrepo-ng PRs? ( and sorry about contacting you over your private gmail )

PR #1810, opened 2021-03-30
Fix rdcman.sls downloads

TheBigBear:


TheBigBear:
@twangboy I have put in the waybackmachine links, but commented them out and only converted this old pkg so it could be cleanly listed and removed from salt. You agree?

twangboy (approved):

PR #1811, opened 2021-03-30
Fix activeperl downloads

TheBigBear:


TheBigBear:
@twangboy is this the 'right' solution? to basically allow any installations of this old activeperl pkg to allow to be cleanly listed and or removed by salt but basically stop supporting this for any more recent not quite as outdated versions? not sure the pkg warrants the effort required? not had a single update request in four years or so since these were current ..

PR #1822, opened 2021-04-01
Update url rewrite

keslerm:


keslerm:
It looks like MS pulled the 7.2.1952 installers down.

TheBigBear:
@keslerm  I have updated urlrewrite to latest and archived old releases in such a way that they can still be listed and / or uninstalled. Had to get msi product GUID for that, so the uninstall works without having the msi to hand. 
[ see https://github.com/saltstack/salt-winrepo-ng/pull/1823 ]

TheBigBear:
@keslerm winrepo-ng dos not want to drop off or simply removed old or no longer available pkg release as they would risk to become stuck in an install. They want the old ones to still be listable and if at all possible uninstallable.

keslerm:
@TheBigBear looks good i'll close this one

PR #1830, opened 2021-04-03
Update spybot.sls to ver. 2.7.64

TheBigBear:


spybot.sls
```diff
@@ -6,9 +6,9 @@
 {% endif %}
 # Source: http://www.safer-networking.org/
 spybot:
-  '2.4':
```
twangboy, line 9: If the old version is no longer available, let's at least leave the `uninstall` data for it

twangboy (changes requested):

spybot.sls
TheBigBear: ok, not sure why I did not do this on this one. Have done this now.

twangboy (approved):

PR #1833, opened 2021-04-03
Update strawberryperl to ver. 5.32.1.1

TheBigBear:


TheBigBear:
@twangboy ok, these pre-commit checks are still broken. The actual URLs work just fine. So I guess these are some that @dafyddj did not yet manage to deal with and fix as he looked at issue https://github.com/saltstack/salt-winrepo-ng/issues/1835 
@twangboy I believe this can be merged as it is 'only' the pre-commit checks that fail not an error in the pkg sls per se.

PR #1834, opened 2021-04-03
Fix rakudo star dl link and update to ver. 2020.10

TheBigBear:


TheBigBear:
These are pre-commit check errors. It fails on a 403 which it should allow for and let pass. The downloads work. The website has introduced a 403 redirect in the middle but the pkg works as expected. This can be merged even though the checks fail. [ see issue 1835 ]( https://github.com/saltstack/salt-winrepo-ng/issues/1835) where we are tracking getting the pre-commit checks fixed to accept 403 or maybe introducing a ignore 403 error flag, whichever seems more appropriate.

twangboy (approved):

PR #1839, opened 2021-04-07
Fix patchmypc-free.sls version number

TheBigBear:
this software NEVER installs itself and always updates itself from same Url with no version number being pre-announced or given


TheBigBear:
the pre-commit tests fail, but it is a test failure and not a pkg failure. the dl script gets a 403 and redirects to get latest version. so this PR is working and can be merged.

twangboy (approved):

Issue #1835, opened 2021-04-04
pre-commit URL check fails on 403 - needs to pass them


TheBigBear:
The ```pre-commit``` hook fails on URLs that actually do work. 

Can this be fixed? @dafyddj 

This is branch ```update-rakkudo-star``` on [my own repo here](https://github.com/TheBigBear/salt-winrepo-ng/tree/update-rakkudo-star)

```
---( rakudo-star-no-jit_x86.sls )---
58
PROBLEM HERE (403) : rakudo-star-no-jit-2016.01_x86 -- 1.0.0 -- https://rakudo.org/dl/star/rakudo-star-2016.01-01-win-x86-(no%20JIT).msi 
59
PROBLEM HERE (Bad content type) : rakudo-star-no-jit-2016.01_x86 -- 1.0.0 -- https://rakudo.org/dl/star/rakudo-star-2016.01-01-win-x86-(no%20JIT).msi -- text/plain 
60
---( rakudo-star_x64.sls )---
61
PROBLEM HERE (403) : rakudo-star-2020.05.1_x64 -- 1.0.0 -- https://rakudo.org/dl/star/rakudo-star-2020.05.1-01-win-x86_64-(JIT).msi 
62
PROBLEM HERE (Bad content type) : rakudo-star-2020.05.1_x64 -- 1.0.0 -- https://rakudo.org/dl/star/rakudo-star-2020.05.1-01-win-x86_64-(JIT).msi -- text/plain 
63
PROBLEM HERE (403) : rakudo-star-2020.01_x64 -- 1.0.0 -- https://rakudo.org/dl/star/rakudo-star-2020.01-01-win-x86_64-(JIT).msi 
64
PROBLEM HERE (Bad content type) : rakudo-star-2020.01_x64 -- 1.0.0 -- https://rakudo.org/dl/star/rakudo-star-2020.01-01-win-x86_64-(JIT).msi -- text/plain 
65
PROBLEM HERE (403) : rakudo-star-2019.03_x64 -- 1.0.0 -- https://rakudo.org/dl/star/rakudo-star-2019.03-01-win-x86_64-(JIT).msi 
66
PROBLEM HERE (Bad content type) : rakudo-star-2019.03_x64 -- 1.0.0 -- https://rakudo.org/dl/star/rakudo-star-2019.03-01-win-x86_64-(JIT).msi -- text/plain 
67
[...]

PR #1843, opened 2021-04-07
Update ossec-agent.sls to ver.  3.6.0 14954

TheBigBear:


TheBigBear:
This is a pre-commit URL check failure. The current code fails on 403 ( redirect ) which are actually perfectly working. So this can be merged as-is. even though the checks appear to have  failed.  [see issue 1835](https://github.com/saltstack/salt-winrepo-ng/issues/1835)

twangboy:
@dafyddj Here's another one that is failing the url test with valid (seeming) urls.

twangboy (approved):

PR #1847, opened 2021-04-08
Update malwarebytes.sls to ver. 4.3.0.210

TheBigBear:


malwarebytes.sls
```diff
@@ -6,12 +6,25 @@
...
+#    installer: 'https://data-cdn.mbamupdates.com/web/mb4-setup-consumer/MBSetup.exe'
+#    install_flags: '/SP- /VERYSILENT /SUPPRESSMSGBOXES /NORESTART'
+#    uninstaller: '{{ PROGRAM_FILES }}\Malwarebytes Anti-Malware\unins000.exe'
+#    uninstall_flags: '/SP- /VERYSILENT /SUPPRESSMSGBOXES /NORESTART'
+#    msiexec: False
+#    locale: en_US
+#    reboot: False
+#  {% endfor %}
+#
+  '4.3.0.210':
```
twangboy, line 21: Looks like they've migrated to a "latest" paradigm as well, no longer offering older versions. May need to use `latest` as the version as we do with chrome.

I still found a url for 3.8 though: https://data-cdn.mbamupdates.com/web/mb3-setup-consumer/mb3-setup-consumer-3.8.3.2965-1.0.629-1.0.13137.exe

twangboy (changes requested):

malwarebytes.sls
TheBigBear: @twangboy but how do I deal with 'latest' IF the version umber is part of the 'full_name' that win pkg tries to match against?

twangboy: I don't know... perhaps it won't work in this case then.

twangboy (approved):

PR #1848, opened 2021-04-08
Update ntp.sls to ver. 4.2.8p15-v2

TheBigBear:


TheBigBear:
The checks failing on this one is also the same now known pre-commit checking error. It fails 403 web redirects, that actually work just fine. So this can be merged and the pre-commit hook needs looking at. [see issue 1835](https://github.com/saltstack/salt-winrepo-ng/issues/1835)

twangboy (approved):

twangboy:
@dafyddj Could you take a look at the test failures here... the urls look valid

PR #1852, opened 2021-04-08
Update bandizip.sls to ver. 6.26 + 7.16

TheBigBear:


bandizip.sls
```diff
@@ -1,14 +1,27 @@
...
-{% set version = '6.21' %}
-
+{% set source_path_v6 = 'http://ca-dl.bandisoft.com/bandizip' %}
+{% set source_path_v7 = 'http://dl.bandisoft.com/bandizip.std' %}
+{% set version6 = '6.26' %}
+{% set version7 = '7.16' %}
+#
 bandizip:
-  '{{ version }}':
+  '{{ version6 }}':
```
twangboy, line 8: If the host only provides the latest version, then perhaps we treat this like chrome and use "latest" as the version.

twangboy (changes requested):

bandizip.sls
TheBigBear: ok, done that.

twangboy:
I'm not sure what to do with this one now that we're using latest. We have Bandizip, Bandizip6 and Bandizip7... all with the same full_name. Not sure how salt will know which one is installed.

twangboy (approved):

twangboy (approved):

PR #1858, opened 2021-04-28
Updated tortoise-hg to current 2021 version

gracinet:
The previously listed versions were out of support for years.

Context: I am a Mercurial core contributor, currently working to provide Windows CI resources for the Mercurial and TortoiseHg projects, and been using Salt on Linux for years (with some minor contribs). 

Being able to manage these machines with Salt will be a drastic simplification to me so I'd like to thank you for making salt_winrepo_ng.

The new definition works for me on an amd64 (x64) VM with Windows Server 2019.

I didn't have the occasion to try it on x86, that's why I kept a general `else` Jinja clause instead of comparing to the actual architecture.

### DIFF
--- a/tortoise-hg.sls
+++ b/tortoise-hg.sls
@@ -1,5 +1,3 @@
-# Source: http://tortoisehg.bitbucket.org/
-
 {% if grains['cpuarch'] == 'AMD64' %}
 {% set msi_arch = 'x64' %}
 {% else %}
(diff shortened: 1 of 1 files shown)


tortoise-hg.sls
```diff
@@ -1,34 +1,16 @@
-# Source: http://tortoisehg.bitbucket.org/
+{% if grains['cpuarch'] == 'AMD64' %}
+{% set msi_arch = 'x64' %}
+{% else %}
+{% set msi_arch = 'x86' %}
+{% endif %}
+
 tortoise-hg:
```
twangboy, line 7: We still need to keep the older versions around so that they can be uninstalled.

twangboy (changes requested):

tortoise-hg.sls
gracinet: Didn't think of that, of course, makes sense. I pushed a correction.

Sadly the old 3.3.0 and 3.6.2 are not currently available from the official download site, so we have to hope that people that kept the MSIs.

I've asked the TortoiseHg maintainers.

gracinet:
Also fixed a leftover `x64` in first commit

TheBigBear:
In cases like, either because it had CVEs or old versions that are no longer available,  then just leave the uninstall related info and remove the install related info. 

( look at npp.sls and how it deals with versions prior to ver. 7.6.6 that are no longer publicly available. Or you could leave the installer URL in and do a skip open the test with the
"skip_urltest: True” line for those.

gracinet:
@TheBigBear thanks for the review and the tips!

I've disabled the install for the old versions.

twangboy (approved):

gracinet:
@twangboy great, many thanks!

PR #1868, opened 2021-06-02
prometheus wmi exporter: add version 0.15 and 0.16

TerraTalpi:
I have added the two new versions 0.15 and 0.16 to the prometheus wmi exporter pkg definition.


TerraTalpi:
I have now fixed issue with the changed field `full_name` starting with version `0.13.0`. This was actually a bug that I just saw right now. The `full_name` before `0.13.0` was `WMI Exporter`, starting with `0.13.0` it has changed to `windows_exporter`.

The full SLS has now been tested by me in our environment, including removing, installing and updating the exporter. Can be merged from my side now.

twangboy (approved):

PR #1869, opened 2021-06-04
Added latest LTS releases for Blender (2.93.0, 2.83.15, and 2.83.14).

ricekab:
This updates the blender.sls definition to include the latest LTS releases (2.93.0 and 2.83.15).

This only defines 64-bit installers as blender doesn't support 32-bit for Windows since version 2.80 (See: https://developer.blender.org/T67184)

### DIFF
--- a/blender.sls
+++ b/blender.sls
@@ -15,3 +15,17 @@ blender:
     msiexec: True
     locale: en_US
     reboot: False
+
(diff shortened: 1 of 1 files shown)


ricekab:
Last commit fixed uninstall procedure (the full_name and version didn't match)

twangboy (approved):

PR #1872, opened 2021-06-16
Create jdk11.sls

sharkbruhaha:


sharkbruhaha:
Not sure what the check about commits wants here... This is a simple add, so not much in the way of commits.

PR #1896, opened 2021-10-30
Update winrar.sls , npp.sls and added npp_x64.sls, updated handbrake.sls

xzenor:
Version(s) 6.x added


xzenor:
updated npp versions and added x64 edition

xzenor:
If my pull request is being ignored because I did something wrong, please let me know so I can fix it.
I'm still trying to get the hang of github.

twangboy (approved):

twangboy:
In the future, please break out each product into its own PR. Thanks for the code!

twangboy:
Looks like the pre-commit is failing. Winrar has released a new beta... the old one is no longer available. There's probably an argument for not including beta releases in package definitions.

xzenor:
> In the future, please break out each product into its own PR. Thanks for the code!

Yeah I figured that out by now. I wanted to help out but GitHub was still fairly new to me and you now had to experience my inexperience.. Sorry about that. I'll split them per app next time.

> Looks like the pre-commit is failing. Winrar has released a new beta... the old one is no longer available. There's probably an argument for not including beta releases in package definitions.  

Fixed it. I agree with not including the beta releases anymore.

TheBigBear:
@twangboy @dafyddj sorry @xzenor I had asked questions about the way forward with npp and going from v 7 to v 8 and how to best deal with the now available 64-bit version. And I think I really liked one of @dafyddj's suggestions. I will link and mark his suggestion to this PR patch of yours. Maybe you should split off the npp away from this PR as it might best be handled a different way? I was still waiting for comments and a decision from @twangboy but was not actively pursuing it enough. sorry about that everybody.

xzenor:
So, just to keep things neat and doing things the way they should be... Would it be better if I just closed this pull request and start creating separate pull requests per application?  

That way some pull requests can be merged while others (like npp) can take some extra time.

TheBigBear:
@xzenor as painful as this seems, I think that might be the better option, or instead of deleting it, you could simply convert it into a DRAFT PR for now, so it won't be considered for merging, but might be a bit easier to read the history and why it is where it is. (see https://github.blog/2019-02-14-introducing-draft-pull-requests/ )

xzenor:
@TheBigBear nah, it's fine. I'd like to do it the right way. Not your fault that I'm a Github n00b.

It's not that much work and it's also some more practice for me.

PR #1900, opened 2021-11-24
Update winrar.sls

xzenor:
Added v6 versions
Because the links to older betas are removed by RarLab, betas are not included.


xzenor:
@TheBigBear Good point. That just leaves 6.02 though.. as anything before 6.02b1 contains CVE-2021-35052 remote code exploit.
And since we're skipping beta versions (for good reasons) there's only 6.02 left.

Might take a bit to figure this out. Thanks for pointing me to an example.

twangboy (approved):

PR #1910, opened 2022-01-29
Update winrar.sls

xzenor:
WinRAR 6.10 was released


xzenor:
Ugh.. Apologies. RarLab changed the naming for the x32 version in 6.10. (Really, who uses x32 windows nowadays?)
It's for the better (more consistent with x64 version) but it does mess up some things if I want to keep old versions available as well.

twangboy (approved):

PR #1932, opened 2022-03-28
Npp x64

xzenor:
The x64 edition of Notepad++
The first x64 edition of Notepad++ is from 2016. I think it's about time to start providing it through in the winrepo now as a choice. Most (Not all) plugins are available in x64 by now and if you want to open files bigger than 4GB you can't use the x32 one anyway.


xzenor:
Okay @TheBigBear , I must admit, I did not expect this. It actually upgraded the 32Bit version to the x64 version.
I tried an older x32 and the latest x32 version and on both tests the x32 one got replaced by the latest x64 one.

I'm not sure what would happen on a 32 bit system though as I don't have any but an arch check should probably be put in there.

twangboy:
We should be explicit and keep this one as `npp_x64` and the 32-bit version as `npp_x86`

PR #1940, opened 2022-08-05
Update malwarebytes.sls

saintdle:
Updated file to use new installer and uninstall methods. 
Issues with versioning of product and release files from official source, so included note about this.

### DIFF
--- a/malwarebytes.sls
+++ b/malwarebytes.sls
@@ -1,25 +1,17 @@
-# just 32-bit x86 installer available
-{% if grains['cpuarch'] == 'AMD64' %}
-    {% set PROGRAM_FILES = "%ProgramFiles(x86)%" %}
-{% else %}
-    {% set PROGRAM_FILES = "%ProgramFiles%" %}
-{% endif %}
 # Source: http://www.malwarebytes.org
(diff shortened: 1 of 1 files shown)


malwarebytes.sls
```diff
@@ -1,30 +1,36 @@
-# just 32-bit x86 installer available
-{% if grains['cpuarch'] == 'AMD64' %}
-    {% set PROGRAM_FILES = "%ProgramFiles(x86)%" %}
-{% else %}
-    {% set PROGRAM_FILES = "%ProgramFiles%" %}
-{% endif %}
 # Source: http://www.malwarebytes.org
+# This is a terrible setup, as Malwarebytes give you a single installer with no version information, nor do they release previous versions to
+# download from their official site. 
+# Use this with caution.
```
twangboy, line 4: Chrome does the same thing. They do not provide previous versions, only an MSI of the latest version. The way we handle this scenario is to use `latest` as the version as we have no way of knowing what the actual version is. For example, look at the Software Definition for Chrome:
https://github.com/saltstack/salt-winrepo-ng/blob/master/chrome.sls

twangboy (changes requested):

twangboy:
Closing due to inactivity

PR #1950, opened 2022-08-29
vlc fix jinja syntax errors removed conflicting IDs

TheBigBear:
why did this pass the tests on last commit?


vlc.sls
```diff
@@ -1,7 +1,7 @@
 # both 32-bit (x86) AND a 64-bit (AMD64) installer available
 {% set PROGRAM_FILES = "%ProgramFiles%" %}
 vlc:
-{% for version in ['3.0.17.4', '3.0.17.3', '3.0.17', '3.0.16', '3.0.15', '3.0.14', '3.0.13', '3.0.12,1', '3.0.12', '3.0.11.1', '3.0.11', '3.0.10', '3.0.9.2', '3.0.8', '3.0.7.1', '3.0.7', '3.0.6', '3.0.4'] %}
+{% for version in ['3.0.17.4', '3.0.17.3', '3.0.17', '3.0.16', '3.0.15', '3.0.14', '3.0.13', '3.0.12,1', '3.0.12', '3.0.11.1', '3.0.11', '3.0.10', '3.0.8', '3.0.6', '3.0.4'] %}
```
twangboy, line 4: These are conflicting with versions that have had their assets removed. I think the problem is that the assets for these haven't been removed. They probably shouldn't be in the list below. That is true for the following which do have assets:
- 3.0.9.2
- 3.0.7.1
- 3.0.7
So, these^^^ should be removed from the list on line 26

twangboy (changes requested):

twangboy (approved):

PR #1951, opened 2022-08-29
patchmypc-free adapt local cache location

TheBigBear:

### DIFF
--- a/patchmypc-free.sls
+++ b/patchmypc-free.sls
@@ -5,7 +5,7 @@ patchmypc-free:
     installer: 'https://patchmypc.net/freeupdater/PatchMyPC.exe'
     install_flags: |
                    ' & cmd /c taskkill /F /IM PatchMyPC.exe &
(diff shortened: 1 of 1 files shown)


twangboy:
The location of the non-binary data is not configurable via the installer. There may be cases where someone upgraded from an old version of salt and did not select to adapt the new paradigm for file location. In that case, the cache would still be in `C:\salt`. So, perhaps there may need to be some logic here. I already merged a few other PRs with this same type of change.

twangboy:
The new installers use a registry entry under `HKLM:\\SOFTWARE\\Salt Project\\Salt`. The key is called `root_dir`. There is a config setting that gets its default value from the registry setting. So, if nothing is set in config, the value will be the registry setting. If it is set in the config, it will be the config value. So, best to get the `root_dir` by calling `config.get root_dir`. That should be available in the jinja parser.

twangboy:
```
{% set install_dir = salt["config.get"]("cachedir") %}
```

twangboy:
Closing due to inactivity

PR #1954, opened 2022-08-29
pycharm-pro adapt local cache location

TheBigBear:


twangboy (approved):

twangboy:
This doesn't need to be conditional. The location of the winrepo cache isn't changing.

PR #1958, opened 2022-10-14
New versions of nxlog

prensing:
Updated versions of nxlog. Previous version 2.10.2150 is no longer available online, so I removed it.

### DIFF
--- a/nxlog.sls
+++ b/nxlog.sls
@@ -1,4 +1,12 @@
 nxlog:
+  '2.10.2100':
+    full_name:  'NXLog-CE'
+    uninstaller: 'https://nxlog.co/system/files/products/files/348/nxlog-ce-2.10.2150.msi'
+    uninstall_flags: '/quiet /norestart'
+    msiexec: True
+    locale: en_US
+    reboot: False
+
   '2.11.2190':
     full_name:  'NXLog-CE'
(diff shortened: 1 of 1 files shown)


twangboy (changes requested):
We need to leave the definition for unsupported/removed version. Just remove the installer portion of the definition (`installer` and `install_flags`). This makes the package remove only. It will allow users who have the old version installed to remove that version.

prensing:
I put the old version back, removing the "installer" variables. However, the uninstall references the specific version of the MSI, which is no longer available for download. Is that going to work?

nxlog.sls
```diff
@@ -1,10 +1,28 @@
 nxlog:
-  '2.10.2150':
+  '2.10.2100':
```
twangboy, line 2: 2150

twangboy (changes requested):
For now we'll leave the uninstaller. In the future we need to use the product GUID for the uninstaller instead of an MSI at a URL. Then if the resource is removed, we can still uninstall.

nxlog.sls
prensing: Sorry about that, I just checked in the change

twangboy (approved):

PR #1991, opened 2023-03-07
Update nextcloud-client.sls . Add new versions

alexanderhaensch:
Nextcloud client development was speeding up. Adding all new versions since 3.4.4


twangboy:
@alexanderhaensch Looks like the URL tests are failing on the x86 downloads. Looks like, as of version `3.6.0` x86 versions are no longer built. Could you modify the Software Definition file so that it doesn't generate an `x86` URL?
You could look at the `git.sls` for an example using `pkg.compare_versions`. You could also create separate version lists like `salt.sls`.
Please reach out if you need any help with this.
Thanks!

twangboy:
Closing this due to activity

PR #2067, opened 2024-01-17
Update npp.sls latest version

lamdaiphong:
As old version 8.5.4 has been reported vulnerabilities security hence required to install higher version.


lamdaiphong:
Am updating another applications as well, will PR later.

PR #2068, opened 2024-01-18
Update npp.sls

lamdaiphong:
version 8.5.4 has been reported security vulnerabilities, update latest version.


npp.sls
```diff
@@ -70,6 +77,9 @@ npp:
 # correctly in `pkg.list_pkgs` and to allow for removal using `pkg.remove`
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   {% for version in [
+    '8.6.2',
+    '8.6.1',
+    '8.6',
```
twangboy, line 82: You don't want these here. There are still assets for them on github.

twangboy (changes requested):

npp.sls
lamdaiphong: I see, indeed don't need the x64 so can removes these.

twangboy (approved):

npp.sls
```diff
@@ -6,6 +6,13 @@
 {% endif %}
 npp:
   {% for version in [
+    '8.6.2',
+    '8.6.1',
+    '8.6',
+    '8.5.8',
+    '8.5.7',
+    '8.5.6',
+    '8.5.5',
```
twangboy, line 15: The URL tests fail on `8.5.5`. It doesn't have a download artifact. You can either not add this version, or you can move it to the bottom list.

twangboy (changes requested):

lamdaiphong (commented):
Removed version 8.5.5

twangboy (approved):

twangboy:
Thanks for getting those changes in!

TheBigBear:
@twangboy is this correct to entirely remove that version? I would have thought of creating a 'with CVE' category and removing its ability to be installed but leaving in the section to be uninstalled. Much like you guys do for the windows salt-minion would be better? No?

twangboy:
That's basically what that list at the bottom is for. Since you're adding a new version, it is not required that you add the CVE version of npp to the bottom list. If this was a version that used to be managed using a salt package definition, then you would need to move it to the bottom list. That bottom list is basically remove only as it does not define an installation parameter.

PR #2075, opened 2024-02-23
Added Blender LTS releases as of 20240223

ricekab:
Latest Blender long term support releases


dafyddj:
Thanks for you contribution.

Can you take a look at the indenting of `'{{ version }}.{{ patch }}':` It should be indented by 2 spaces, not 4.

Check the "Run URL tests" results to make sure that all versions are accounted for in the test.

ricekab:
Thanks for applying the changes! Was a bit busy over the week.

PR #2079, opened 2024-02-27
zoom: add 64-bit installer

### LINKED ISSUE #2053: Zoom installer on Windows defaults to 32-bit version
zoom.sls
{% set url = "https://www.zoom.us/client/latest/ZoomInstallerFull.msi" %}

This url will install 32-bit latest client, overwriting existing 64 client installed on the system
to make it install 64 bit zoom it should be  
{% set url = "https://www.zoom.us/client/latest/ZoomInstallerFull.msi?archType=x64" %}

The problem is that 32 bit client cannot be upgraded - it can only 
be overwritten with 64 bit installer with higher version

Soon (in march 2024) I believe zoom will drop support to 32 bit versions, so why to default to this?

dafyddj:
Resolves #2053

### DIFF
--- a/zoom.sls
+++ b/zoom.sls
@@ -1,10 +1,21 @@
-{% set url = "https://www.zoom.us/client/latest/ZoomInstallerFull.msi" %}
-zoom:
+{% set data = {
+     64: {
+       "pkg": "_x64",
+       "name": " (64-bit)",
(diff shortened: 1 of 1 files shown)


twangboy (approved):

dafyddj:
Thanks you!

PR #2099, opened 2024-03-26
Fix repo path

twangboy:
The `winrepo.update_git_repos` creates a directory named `salt-winrepo-ng_git`. The directory based installations are all looking in `salt-winrepo-ng`. If you try to install any of these they can't find their resources because they're looking in the wrong location.

### DIFF
--- a/auditbeat/init.sls
+++ b/auditbeat/init.sls
@@ -12,8 +12,8 @@ auditbeat:
 {% for version in versions %}
   '{{ version }}':
     full_name: 'Auditbeat'
-    installer: 'salt://win/repo-ng/salt-winrepo-ng/auditbeat/install.cmd'
+    installer: 'salt://win/repo-ng/salt-winrepo-ng/_/auditbeat/install.cmd'
     install_flags: {{ version }}
-    uninstaller: 'salt://win/repo-ng/salt-winrepo-ng/auditbeat/remove.cmd'
+    uninstaller: 'salt://win/repo-ng/salt-winrepo-ng/_/auditbeat/remove.cmd'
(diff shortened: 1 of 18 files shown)


TheBigBear:
What ? ! @twangboy ? It has always been ```salt-winrepo-ng``` why on earth did ```winrepo.update_git_repos``` create such an ugly path? This will add ```.git``` at end of it on git servers and make for a really confusing ( and IMHO 'wrong' and 'ugly' looking path. 
Wouldn't the correct fix be to correct the ```winrepo.update_git_repos``` to create the correct directory in the first place. 
It did the wrong change? No?

twangboy:
The problem is in the gitfs code that `winrepo.update_git_repos` uses. A PR that went in last fall changed some things. On a masterless minion it appended `_git` to all the directories winrepo uses. That's what I was trying to fix here before I realized this is a deeper problem. On a Linux master it creates a subdirectory under the normal directory that is just an underscore (`_`). I have a PR in to fix those, but I don't know what else I'm breaking with it. I don't understand why the changes were made in the first place... what problem they were trying to solve.

https://github.com/saltstack/salt/pull/66283

twangboy:
So, the final fix was to accept the `_` directory that gets created on the master, or when the `update_git_repos` command uses PyGit2 or GitPython. I changed the legacy method (using only git) to create the same directory structure, adding a `_` directory. So, I'll update this PR to follow that directory structure. This should only affect these script-based package definitions that are looking for specific files in the cache.

twangboy:
These packages will work again starting with the 3006.8 and 3007.1 releases.

PR #2106, opened 2024-04-12
update graylog-collector-sidecar to graylog-sidecar

krombel:
Upstream renamed the sidecar from collector-sidecar to graylog-sidecar

=> Update this here accordingly and track the latest releases

### DIFF
--- a/graylog-collector-sidecar.sls
+++ b/graylog-collector-sidecar.sls
@@ -1,14 +1,35 @@
-{% set versions = ['0.1.6', '0.1.5', '0.1.4', '0.1.3', '0.1.2', '0.1.1', '0.1.0'] %}
+{%- load_yaml as versions %}
+- 1.5.0
+- 1.4.0
+- 1.3.0
+- 1.2.0
+- 1.1.0
+- 1.0.2
+- 1.0.1
+- 0.1.8
+- 0.1.6
+- 0.1.5
+- 0.1.4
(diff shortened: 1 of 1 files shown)


dafyddj:
Hi, and thanks for your contribution.

We don't delete old versions in this repo, particularly if they are still available to download, so if you could please revert that change.

The change to the package name is a breaking change, and not desirable in my opinion.

The only real variance that's needed is to `full_name` and that can be handled using `salt["pkg.compare_versions"]` as can be seen for example in `git.sls`.

Please use inline yaml for the version list as can be seen in `git.sls` and `salt-minion.sls`

dafyddj:
Hi,

following on from our conversation at the last working group meeting re. your updates, personally I prefer a DRY (don't repeat yourself) approach as seen in `firefox.sls` but Shane follows a similar style to yours in `salt-minion.sls' so it's a personal choice.

Some minor points but the fields `locale` and `reboot` are non-functional and only to serve to confuse so I would leave these out, and `msiexec: False` is the default so again I would leave out.
To aid readability I would place the latest code block first and the older (versions `<"1.0.0"`) below so it's easier to scan the file.

Thanks

krombel:
@dafyddj I removed the locale, reboot and msiexec flag and moved "newer" to top.
I hope it is fine now

dafyddj (approved):
Thanks

PR #2298, opened 2025-02-08
Update arduino-ide to Version 2.3.4

AlexanderGabriel:
Updated arduino-ide to Version 2.3.4

Tested update from 2.3.2 and fresh installation


dafyddj:
Just a note to say that it's preferable that the version list is complete, and it's best not to skip versions as someone out there may need to uninstall any previous version, so I would suggest adding in `2.3.3` at the same time. (Same PR is fine).

PR #2658, opened 2026-04-19
Update Notepad++ x64 to 8.9.3

AlexanderGabriel:
Update Notepad++ x64 to Version 8.9.3
removed trailing spaces in file


AlexanderGabriel:
And of course tested updating from older version, downgrade to older and uninstallation

